PR #17, opened 2018-05-10
fix: mwc-textfield disabled input

### LINKED ISSUE #14: TextInput demo: Disabled input boxes are targetable and editable using tab
While you cannot click on disabled text boxes I noticed they can be tabbed to and then edited.

romulocintra:
Closes #14 

Missing test for textfields


azakus:
Sorry for the delay, but this is fixed with #297

PR #42, opened 2018-05-21
Bug fix: Forwarding the change event from the input element

DominicWrege:
Hi this is my first pull request. So I made one little change to the switch element to improve it.
Sorry my commit message is a little bit broken. Anyhow I think the change event got lost because of some shadow DOM things.  

**commit message again:** 
> "bug fix: forwarding the change event from the input element and setting the check property value"


DominicWrege:
Hello @googlebot please wake up. I signed it!

frankiefu:
Fixed in https://github.com/material-components/material-components-web-components/pull/258

PR #45, opened 2018-05-23
Fix dependency references and tests.

bicknellr:
- `mwc-fab` is now in its own package.
- `mwc-ripple-surface` doesn't seem to exist as a separate component anymore.
- Adds `async` to a few test functions that were using `await`.

There are still some left over uncaught errors that happen when running the tests that I'm still looking into. They don't cause the tests to fail though, so they must be async.

### DIFF
--- a/test/unit/mwc-checkbox.test.js
+++ b/test/unit/mwc-checkbox.test.js
@@ -41,11 +41,12 @@ test('element._formElement returns the native checkbox element', () => {
   assert.equal(element._formElement.localName, 'input');
 });
 
-test('element._component returns the mdc checkbox component', () => {
+test('element._component returns the mdc checkbox component', async () => {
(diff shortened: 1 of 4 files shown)


test/unit/mwc-checkbox.test.js
```diff
@@ -41,11 +41,12 @@ test('element._formElement returns the native checkbox element', () => {
   assert.equal(element._formElement.localName, 'input');
 });
 
-test('element._component returns the mdc checkbox component', () => {
+test('element._component returns the mdc checkbox component', async () => {
+  await element.componentReady();
```
bicknellr: This is the only functional change - I'm not sure if it's correct but other tests seemed to do this before reading `._component` and adding it here caused this test to stop failing.

PR #47, opened 2018-05-25
Checkbox and iron-form

eskan:
Hi,
this is more a proposal than a fix PR.
The idea is to get this element works with iron-form so i had the **name** attribute and also a property : **_hasIronCheckedElementBehavior** to allow iron-form to handle checkbox checked value.

I'm not sure this is a good way to handle MWC form, may be a transitional way with iron-form until a MWC alternative exists ?


merlinnot:
I'd love to see it working with a native `form`.

eskan:
@merlinnot you're right :D i've changed the title

eskan:
outdated

PR #53, opened 2018-05-30
Extra tests for button, FAB. Add `label` property to FAB.

bicknellr:
- Are the tests too brittle by they way they look into the internals to find elements?
- I added `label` as a property of fab because users will want to label the button in FABs with text that's more descriptive than just the icon name.
- Tests for the FAB are split into a different file.

### DIFF
--- a/packages/fab/mwc-fab.js
+++ b/packages/fab/mwc-fab.js
@@ -36,7 +36,7 @@ export class Fab extends LitElement {
     this.icon = '';
     this.mini = false;
     this.exited = false;
-    this.label = undefined;
+    this.label = '';
   }
 
   _createRoot() {
--- a/packages/fab/mwc-fab.js
+++ b/packages/fab/mwc-fab.js
@@ -27,6 +27,7 @@ export class Fab extends LitElement {
       exited: Boolean,
       disabled: Boolean,
       icon: String,
+      label: String,
     };
   }
 
@@ -35,6 +36,7 @@ export class Fab extends LitElement {
     this.icon = '';
     this.mini = false;
     this.exited = false;
+    this.label = undefined;
   }
 
   _createRoot() {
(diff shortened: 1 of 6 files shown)


packages/fab/mwc-fab.js
```diff
@@ -51,14 +53,14 @@ export class Fab extends LitElement {
...
-  _render({icon, mini, exited, disabled}) {
+  _render({icon, mini, exited, disabled, label}) {
     const hostClasses = c$({
       'mdc-fab--mini': mini,
       'mdc-fab--exited': exited,
     });
     return html`
       ${this._renderStyle()}
-      <button class$="mdc-fab ${hostClasses}" disabled?="${disabled}" aria-label$="${icon}">
+      <button class$="mdc-fab ${hostClasses}" disabled?="${disabled}" aria-label$="${label || icon}">
```
sorvell: Does it make sense to be the button `title` as well?

test/unit/mwc-button.test.js
```diff
@@ -35,17 +37,29 @@ test('initializes as an mwc-button', () => {
...
+  element.disabled = true;
+  await afterNextRender();
+  const button = element.shadowRoot.querySelector('button');
+  assert.equal(button.hasAttribute('disabled'), true);
 
-beforeEach(() => {
-  element = document.createElement('mwc-fab');
-  document.body.appendChild(element);
+  element.disabled = false;
+  await afterNextRender();
```
sorvell: So, LitElement provides a `renderComplete` Promise that we should use for this. 

NOTE: It's a little tricky because `ComponentElement` provides `componentReady` (as a method) and probably should instead override `renderComplete`. Button just extends LitElement so that's not an issue here.

test/unit/mwc-fab.test.js
```diff
@@ -0,0 +1,89 @@
...
+  document.body.removeChild(element);
+});
+
+test('initializes as an mwc-fab', () => {
+  assert.instanceOf(element, Fab);
+});
+
+test('get/set disabled updates the disabled property on the native button element', async () => {
+  element.disabled = true;
+  await afterNextRender();
```
sorvell: Should also use `element.renderComplete`

test/unit/mwc-button.test.js
bicknellr: Thanks for pointing that out, seems much cleaner to have this signal explicitly come from the element.

packages/fab/mwc-fab.js
```diff
@@ -35,6 +36,7 @@ export class Fab extends LitElement {
     this.icon = '';
     this.mini = false;
     this.exited = false;
+    this.label = undefined;
```
sorvell: nit: why not empty string?

packages/fab/mwc-fab.js
sorvell: ok

packages/fab/mwc-fab.js
bicknellr: Fixed.

Issue #62, opened 2018-06-07
Missing textarea


eskan:
Hi,
i would like to help implement the missing textarea component wich is in textfield in vanilla. There's many way to achieve this :
- add a textarea attribute (and rows and col)
- create an mdc-textfield-textarea or mdc-textarea
- other
Do you have think about this already ?
regards

sorvell:
That's great. It probably makes sense to create `mwc-textarea` and make it a subclass of `mwc-textfield`. It might be necessary to add some additional extension points in `mwc-textfield` to make this convenient. There's some info about supported textarea rendering here https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield#textarea.

eskan:
related PR was outdated and closed

elf-pavlik:
https://material.io/design/components/text-fields.html#input-types has an example of ***Multi-line text fields** grow to accommodate multiple lines of text*

I recall polymer paper elements had such functionality for text area. Should autogrow work with textfield or textarea (or both)?

![ezgif-3-a253e84ead3d](https://user-images.githubusercontent.com/876431/47178928-3ea14b80-d2e2-11e8-904e-9a97e426f8fb.gif)

PR #64, opened 2018-06-12
add textarea component

eskan:
Hi,
this is a proposal to add mwc-textarea component. I put it on a different directory to allow scripts to build it. I've also created/updated related files (demos, test).
What is missing : 
- there some missing events that i prefer to wait for this to be merged as textarea is a subclass of texfield : 
https://github.com/material-components/material-components-web-components/pull/46
- components properties and events needs to be documented 
- there are differences in behavior with the vanilla version that must be issued like helper text is always visible in vanilla.
Regards,
Cyrille


eskan:
related to https://github.com/material-components/material-components-web-components/issues/62

eskan:
outdated

PR #35, opened 2018-05-14
Update Button readme

DaniPani:
There was `icon` instead of `button` and I corrected it

PR #89, opened 2018-08-09
Fix typo in mwc-button's README.md

lozandier:


lozandier:
@andrealeardini Acknowledged, will also correct that.

elf-pavlik:
looks like duplicate of #35

aomarks:
Thanks! Fixed by slightly earlier https://github.com/material-components/material-components-web-components/pull/35.

PR #101, opened 2018-08-27
Implement extended state for FAB

zevisert:
Updated the `@material/fab` dependency and added a new property to track the `mdc-fab--extended` class.

Also added a fallback for when `extended` is true, but no `label` exists, as the Material Design Guidelines states "[An extended FAB can’t have an icon without a text label.](https://material.io/design/components/buttons-floating-action-button.html#extended-fab)".


azakus:
Sorry for the delay, but this is fixed with #111

PR #109, opened 2018-09-20
First crazy attempt at typescript conversion

### LINKED ISSUE #99: Author elements in TypeScript

azakus:
Look at switch and base packages

Move sass building into scripts

Fixes #99 

**DO NOT MERGE** until all elements are converted


merlinnot:
Hello, just wanted to express an appreciation of this PR as a member of the community :) It will make contributing so much easier and it has an added benefit for all of the folks out there already using TS.

packages/base/src/mwc-base-element.ts
```diff
@@ -0,0 +1,84 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import {LitElement} from '@polymer/lit-element/lit-element.js';
+import {TemplateResult} from 'lit-html';
+export {TemplateResult};
+export {html, property, customElement, query} from '@polymer/lit-element/lit-element.js';
```
sorvell: `export *` ?

packages/base/src/mwc-base-element.ts
```diff
@@ -0,0 +1,84 @@
...
+export {classMap} from 'lit-html/directives/classMap.js';
+export {observer} from './observer.js';
+
+export abstract class MWCBaseElement extends LitElement {
```
sorvell: How about just `BaseElement`. I don't think we need `MWC` on any of the class names.

packages/base/src/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Return an element assigned to a given slot that matches the given selector
+ */
+export function findMatchingAssignedElement(slot: HTMLSlotElement, selector: string) {
```
sorvell: Let's change this to `findAssignedElement` (can just add a TODO for now)

packages/switch/src/mwc-switch.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  @query('.mdc-switch')
+  protected mdcRoot!: HTMLElement;
+
+  @query('input')
+  protected formElement!: HTMLInputElement;
+
+  protected mdcFoundation!: MDCSwitchFoundation;
+
+  protected _boundHandler = (e: Event) => {
```
sorvell: if protected then no leading _. What about `_clickHandler` and making this `private`?

packages/base/src/mwc-base-element.ts
justinfagnani: Also, you can shorten this to `'@polymer/lit-element'`

packages/base/src/mwc-base-element.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Create the adapter for the `mdcFoundation`.
+   *
+   * When overriding, use the following pattern:
+   *
+   * `createAdapter() {return {...super.createAdapter(), () => {}}}`
+   *
+   */
+  protected createAdapter() {
```
justinfagnani: When mdc converts to TypeScript, are this and `createFoundation` going to be parameterized by adapter/foundation type?

packages/switch/src/mwc-switch.ts
```diff
@@ -0,0 +1,88 @@
...
+limitations under the License.
+*/
+import {FormElement, html, property, observer, query, customElement} from '@material/mwc-base/form-element.js';
+import {style} from './mwc-switch-css.js';
+import MDCSwitchFoundation from '@material/switch/foundation.js';
+
+@customElement('mwc-switch' as any)
+export class Switch extends FormElement {
+  @property({type: Boolean})
+  @observer(function(this: Switch, value: boolean) {
```
justinfagnani: Another syntax option here is to add a property to the PropertyOptions object (by overriding `createProperty`):

```ts
@property({
  type: Boolean,
  onChange(this: Switch, value: boolean) {
    this.mdcFoundation.setChecked(value);
  },
})
checked = false;
```

packages/switch/src/mwc-switch.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.checked = this.formElement.checked;
+  }
+
+  protected get mdcFoundationClass() {
+    return MDCSwitchFoundation;
+  }
+
+  protected createAdapter() {
+    return {
+      ...super.createAdapter(),
```
justinfagnani: Not related to TS, and I _think_ we might have talked about this already, but using classes would reduce memory by using prototypes and not instantiating each method as a separate function object.

Maybe something like:

```ts
  static protected AdapterClass = class extends super.AdapterClass {
    setNativeControlChecked: (checked: boolean) => {
      this.element.formElement.checked = checked;
    }
  };
```

tsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
+    "module": "es2015",
+    "moduleResolution": "node",
+    "lib": ["es2017", "dom", "dom.iterable"],
+    "declaration": true,
+    "sourceMap": true,
+    "inlineSources": true,
+    "strictFunctionTypes": true,
```
justinfagnani: I'd recommend `"strict": true`

packages/base/src/mwc-base-element.ts
azakus: Yeah I'd like to do that.

packages/switch/src/mwc-switch.ts
azakus: @sorvell WDYT?

packages/switch/src/mwc-switch.ts
azakus: Sounds good to me

tsconfig.json
azakus: I think I had trouble with the implicit any's from importing the MDC Foundations, but I'll give it a shot.

packages/switch/src/mwc-switch.ts
azakus: Worked it out by moving the anonymous class outside of the component class and removing `protected` from the properties accessed by the adapter

tsconfig.json
azakus: I can turn `strict: true` on, but I also have to disable `noImplicitUseStrict`.

Leaving `noImplicitAny` on, otherwise I'd have to provide typings for all the Foundations.

packages/base/src/base-element.ts
```diff
@@ -0,0 +1,97 @@
...
+   * Root element for MDC Foundation usage.
+   *
+   * Define in your component with the `@query` decorator
+   */
+  abstract mdcRoot: HTMLElement;
+
+  /**
+   * Return the foundation class for this component
+   */
+  protected abstract get mdcFoundationClass(): any;
```
rictic: prefer `unknown` over `any`

`any` is contagious. most operations on an `any` produces an `any`, and all operations are allowed, so you can have code far removed from the initial `any` that looks like it's type checked but isn't. this has caused a few bugs in the tools org

in my experience, `unknown` works just as well, occasionally requiring a little more work, but it's way safer by default

packages/switch/src/mwc-switch.ts
sorvell: Perhaps slightly nicer, but this way it's clearly a separate thing (since it's not part of LitElement).... meh.

packages/base/src/form-element.ts
```diff
@@ -0,0 +1,44 @@
...
+import {BaseElement} from './base-element.js';
+export * from './base-element.js';
+
+export abstract class FormElement extends BaseElement {
+  /**
+   * Form-capable element in the component ShadowRoot.
+   *
+   * Define in your component with the `@query` decorator
+   */
+  protected abstract formElement: HTMLInputElement;
```
justinfagnani: HTMLTextAreaElement doesn't inherit from HTMLInputElement: https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement

packages/switch/src/mwc-switch.ts
```diff
@@ -0,0 +1,90 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import {FormElement, html, property, observer, query, customElement, BaseAdapter} from '@material/mwc-base/form-element';
+import {style} from './mwc-switch-css';
+import MDCSwitchFoundation from '@material/switch/foundation';
+
+export class SwitchAdapter extends BaseAdapter {
+  constructor(readonly element: Switch) {
```
justinfagnani: do you need the constructor? I think you can replace it with a type parameter on the base class for the type of `element`.

packages/switch/src/mwc-switch.ts
```diff
@@ -0,0 +1,90 @@
...
+  protected mdcFoundation!: MDCSwitchFoundation;
+
+  private _changeHandler(e: Event) {
+    this.mdcFoundation.handleChange(e);
+    // catch "click" event and sync properties
+    this.checked = this.formElement.checked;
+  }
+
+  protected get mdcFoundationClass() {
+    return MDCSwitchFoundation;
```
justinfagnani: why is AdapterClass a static property and this an instance getter?

packages/switch/src/mwc-switch.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  @query('.mdc-switch')
+  mdcRoot!: HTMLElement;
+
+  @query('input')
+  formElement!: HTMLInputElement;
+
+  protected mdcFoundation!: MDCSwitchFoundation;
+
+  private _changeHandler(e: Event) {
```
sorvell: Change to `_changeHandler = (e: Event) => {`

otherwise you lose the binding.

azakus: Oh yeah

packages/switch/src/mwc-switch.ts
azakus: So is that a no?

packages/base/src/observer.ts
```diff
@@ -0,0 +1,45 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import {PropertyValues} from '@polymer/lit-element/lib/updating-element.js';
+
+export interface Observer {
+  (value: any, old: any): void;
```
43081j: can this be `unknown`? to ensure where it is used, proper type guards are used etc

packages/switch/src/mwc-switch.ts
azakus: I do need the element assigned somewhere though

packages/base/src/observer.ts
azakus: Hmm, I get type errors trying to define the value type in the `switch` observer if these are `unknown`.

```
packages/switch/src/mwc-switch.ts:33:13 - error TS2345: Argument of type '(this: Switch, value: boolean) => void' is not assignable to parameter of type 'Observer'.
  Types of parameters 'value' and 'value' are incompatible.
    Type 'unknown' is not assignable to type 'boolean'.

33   @observer(function(this: Switch, value: boolean) {
               ~~~~~~~~
```

packages/base/src/base-element.ts
azakus: Working with `unknown` seems like a giant PITA. It might just be easier to describe an interface for the foundations

azakus:
Ok, it looks like the foundations actually expect a bag of bounded functions as adapters, so we're going to move back to that structure and file an issue.

We'll also add light typings for the Foundation methods we call to get some type checking and spell-check safety.

packages/checkbox/src/mwc-checkbox.ts
```diff
@@ -0,0 +1,122 @@
...
+  handleAnimationEnd(): void;
+  handleChange(): void
+}
+
+export declare var CheckboxFoundation: {
+  prototype: CheckboxFoundation;
+  new (adapter: Adapter): CheckboxFoundation;
+}
+
+@customElement('mwc-checkbox' as any)
```
43081j: should this cast be here? (`as any`)

azakus: Yes, we're working around https://github.com/Microsoft/TypeScript/issues/27171.
I've filed https://github.com/material-components/material-components-web-components/issues/114 to remove the cast when the upstream bug is fixed.

43081j: ah good to know 👍

packages/base/src/base-element.ts
```diff
@@ -0,0 +1,96 @@
...
+  init(): void;
+  destroy(): void;
+}
+
+export declare var Foundation: {
+  prototype: Foundation;
+  new(adapter: Adapter): Foundation;
+}
+
+export abstract class BaseElement extends LitElement {
```
kenchris: So this BaseElement is only supposed to be used by the FormElement and all other elements should inherit directly from LitElement? maybe deserves a comment

packages/checkbox/src/mwc-checkbox.ts
```diff
@@ -0,0 +1,122 @@
...
+  protected mdcFoundation!: CheckboxFoundation;
+
+  renderStyle() {
+    return style;
+  }
+
+  protected createAdapter(): Adapter {
+    return {
+      ...super.createAdapter(),
+      getNativeControl: () => {
```
kenchris: why not just `getNativeControl: () => this.formElement;`

packages/fab/src/mwc-fab.ts
```diff
@@ -0,0 +1,81 @@
...
+  exited = false;
+
+  @property({type: Boolean})
+  disabled = false;
+
+  @property({type: Boolean})
+  extended = false;
+
+  @property({type: Boolean})
+  showIconAtEnd = false;
```
kenchris: won't this look weird as attribute? `showiconatend` - that is really hard to read

packages/tab-bar/src/mwc-tab-bar.ts
```diff
@@ -0,0 +1,184 @@
...
+      },
+      getIndexOfTab: (tabToFind: Tab) => this._getTabs().indexOf(tabToFind),
+      getTabListLength: () => this._getTabs().length,
+      notifyTabActivated: (index: number) => {
+        // Synchronize the tabs `activeIndex` to the foundation.
+        // This is needed when a tab is changed via a click, for example.
+        this.activeIndex = index;
+        this.dispatchEvent(
+          new CustomEvent(MDCTabBarFoundation.strings.TAB_ACTIVATED_EVENT,
+            {detail: {index}, bubbles: true, cancelable: true}))
```
43081j: do events like this need to be `composed: true` too? they will stop at the nearest shadow boundary otherwise (though you may want that?)

packages/fab/src/mwc-fab.ts
43081j: This isn't reflected by default so shouldn't be a problem for now

43081j:
Would it be worth making most or all of the `@query` properties private/protected?

Most of them seem to be used internally for convenience rather than to be exposed to users of the component. The same applies to the mdcRoot, or is there a reason we keep them public?

azakus:
> Would it be worth making most or all of the `@query` properties private/protected?
> 
> Most of them seem to be used internally for convenience rather than to be exposed to users of the component. The same applies to the mdcRoot, or is there a reason we keep them public?

That's a good question, I've opened #130 to track that

PR #111, opened 2018-09-21
Buttons typescript

sorvell:
Updates
* FAB
* Button
* Icon


packages/ripple/src/ripple-directive.ts
```diff
@@ -31,69 +31,91 @@ export interface RippleOptions {
...
+    registerResizeHandler: (handler: Handler) =>
+      window.addEventListener('resize', handler),
+    deregisterResizeHandler: (handler: Handler) =>
+      window.removeEventListener('resize', handler),
+    updateCssVariable: (varName: string, value: string) =>
+      surfaceNode.style.setProperty(varName, value),
+    computeBoundingRect: () => interactionNode.getBoundingClientRect(),
+    getWindowPageOffset: () => ({x: window.pageXOffset, y: window.pageYOffset}),
+  };
+  const rippleFoundation = new MDCRippleFoundation(adapter);
```
azakus: Can you make a Foundation interface for MDCRippleFoundation like in `mwc-switch`?

azakus:
```ts
const rippleFoundation: RippleFoundation = new MDCRippleFoundation(adapter);
```

azakus:
LGTM after fixing that nit.

PR #128, opened 2018-10-03
use renderComplete

eskan:
afterNextRender have been remove from utils.js in favor of element renderComplete


azakus:
This code isn't going to be used for much longer, but I'll merge the PR.

PR #135, opened 2018-10-05
Update to lit-element 0.6.2

### LINKED ISSUE #132: Package is trying to require "lit-html" without it being listed in its dependencies
The mwc-fab web component depends on [lit-html](https://github.com/Polymer/lit-html) package but this isn't listed in the ```package.json```.

The same problem exists with other material web components too.

https://github.com/material-components/material-components-web-components/blob/4cf9df3e498b116886e51087dd37d1353ed2d3be/packages/fab/src/mwc-fab.ts#L18

https://github.com/material-components/material-components-web-components/blob/4cf9df3e498b116886e51087dd37d1353ed2d3be/packages/fab/package.json#L15-L21

It includes [@polymer/lit-element](https://github.com/Polymer/lit-element) which depends on lit-html, so this shouldn't cause too many problems at the moment, but is already causing problems with at least [Yarn Plug & Play](https://github.com/yarnpkg/rfcs/files/2378943/Plugnplay.pdf).

```Error: Package "@material/mwc-fab@0.2.1" (via "/home/ryanwhite04/.cache/yarn/v3/npm-@material-mwc-fab-0.2.1-24bcfaf5885d58c319a09d751ad1d83d7ea48545/node_modules/@material/mwc-fab/mwc-fab.js") is trying to require the package "lit-html" (via "lit-html/directives/classMap.js") without it being listed in its dependencies (@material/fab, @material/mwc-base, @material/mwc-icon, @material/mwc-ripple, @polymer/lit-element, @material/mwc-fab)```

azakus:
Also make explicit dependencies on lit-html where needed

Fixes #132

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -8,7 +8,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@polymer/lit-element": "^0.6.1"
+    "@polymer/lit-element": "^0.6.2",
+    "lit-html": "^0.12.0"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/button/package.json
+++ b/packages/button/package.json
@@ -17,7 +17,8 @@
     "@material/mwc-base": "^0.3.0",
     "@material/mwc-icon": "^0.3.0",
     "@material/mwc-ripple": "^0.3.0",
-    "@polymer/lit-element": "^0.6.1"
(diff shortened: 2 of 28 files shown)


azakus:
Also had to fix `ripple()` directive to handle a breaking change in lit-html `NodePart`.

PR #141, opened 2018-10-09
Use listener functions on the class

### LINKED ISSUE #136: Change event handlers to class methods
lit-element 0.6.2+ should allow for event handlers to be class methods instead of bound instance functions.

### LINKED ISSUE #149: mwc-checkbox console logs on every render
It seems that version 0.3.1 logs to console this.checked on every render.

https://github.com/material-components/material-components-web-components/blob/master/packages/checkbox/src/mwc-checkbox.ts#L94

azakus:
Add `@eventOptions({passive: true})` for interation handlers in
tab-scroller

Fixes #136
Fixes #149

### DIFF
--- a/packages/radio/src/mwc-radio.ts
+++ b/packages/radio/src/mwc-radio.ts
@@ -118,20 +118,20 @@ export class Radio extends FormElement {
     };
   }
 
-  protected changeHandler() {
+  private _changeHandler() {
     this.checked = this.formElement.checked;
     if (this._selectionController) {
       this._selectionController.update(this);
     }
   }
 
-  protected focusHandler() {
+  private _focusHandler() {
     if (this._selectionController) {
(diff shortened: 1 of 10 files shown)


packages/radio/src/mwc-radio.ts
```diff
@@ -118,20 +118,20 @@ export class Radio extends FormElement {
     };
   }
 
-  private _changeHandler = () => {
+  protected changeHandler = () => {
```
balloob: They can be instance methods instead of bound: `protected changeHandler() {`

azakus:
@balloob Thanks, found a few more as well 😀

azakus:
Looks like this PR also enables the demos for the tab, top-app-bar, and linear-progress.

PR #145, opened 2018-10-10
Update slider and snackbar to mdc v0.40 and typescript

frankiefu:


packages/linear-progress/src/mwc-linear-progress.ts
```diff
@@ -77,16 +77,16 @@ export class LinearProgress extends BaseElement {
   @observer(function(this: LinearProgress, value: boolean) {
     this.mdcFoundation.setReverse(value);
   })
-  reverse = 0;
+  reverse = false;
```
sorvell: Let's make all the `type: Boolean` properties also `reflect: true`. This is the rule on which we've settled.

packages/slider/src/mwc-slider.ts
```diff
@@ -0,0 +1,208 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import {FormElement, html, property, observer, query, customElement, Adapter, Foundation} from '@material/mwc-base/form-element.js';
+import {classMap} from 'lit-html/directives/classMap.js';
+import {repeat} from 'lit-html/directives/repeat.js';
+import {style} from './mwc-slider-css.js';
+import MDCSliderFoundation from '@material/slider/foundation.js';
+
+const {
```
sorvell: Since these values must also be in the HTML here, I'm not sure of the value of using these constants.

packages/slider/src/mwc-slider.ts
```diff
@@ -0,0 +1,208 @@
...
+  @property({type: Boolean})
+  @observer(function(this: Slider, value: boolean) {
+    this.mdcFoundation.setDisabled(value);
+  })
+  disabled = false;
+
+  @property({type: Boolean})
+  discrete = false;
+
+  @property({type: Boolean})
```
sorvell: Make boolean properties reflect.

packages/slider/src/mwc-slider.ts
```diff
@@ -0,0 +1,208 @@
...
+        aria-valuemin="${min}" aria-valuemax="${max}" aria-valuenow="${value}"
+        aria-disabled="${disabled}" data-step="${step}">
+      <div class="mdc-slider__track-container">
+        <div class="mdc-slider__track"></div>
+        ${discrete && markers ? html`<div class="mdc-slider__track-marker-container">
+          ${repeat(new Array(_numMarkers), () => html`<div class="mdc-slider__track-marker"></div>`)}
+        </div>` : ''}
+      </div>
+      <div class="mdc-slider__thumb-container">
+        ${discrete ? html`<div class="mdc-slider__pin">
```
sorvell: Add TODO to use `when` directive when is supports optional `else`.

packages/slider/src/mwc-slider.ts
```diff
@@ -0,0 +1,208 @@
...
+        document.body.removeEventListener(type, handler),
+      registerResizeHandler: (handler: EventListener) =>
+        window.addEventListener('resize', handler),
+      deregisterResizeHandler: (handler: EventListener) =>
+        window.removeEventListener('resize', handler),
+      notifyInput: () => {
+        this.value = this.mdcFoundation.getValue();
+        this.dispatchEvent(new CustomEvent(INPUT_EVENT, {detail: this, bubbles: true, cancelable: true}));
+      },
+      notifyChange: () => {
```
sorvell: Verify that this does not double-notify since setting `this.value` result in setting `setValue` on the foundation (which may cause this notify to go again).

PR #146, opened 2018-10-10
remove imports of afterNextRender (unused)

elf-pavlik:
utils.js doesn't export or even implement that function any more, trying to import it just causes errors


elf-pavlik:
this PR just removes importing function which never gets used, reviewing it should take 1min

Issue #140, opened 2018-10-09
Split components into "base" class for functionality, and "MWC" class for styling


azakus:
Splitting the component classes apart this way will help with theming. If someone wants to theme an MWC component differently, or override part of the theming, they can extend the base class and only have to implement `renderStyle()`

The base class must not register itself as a custom element, and leave that up to the extending class.

Example:

mwc-switch-base.ts:
```ts
export class SwitchBase extends BaseComponent {
  renderStyle() {}
  render() {
    return html`${this.renderStyle()}...`;
  }
}
```

mwc-switch.ts:
```ts
import {SwitchBase} from './mwc-switch-base.js';
import {style} from './mwc-switch-css.js';

@customElement('mwc-switch')
export class Switch extends SwitchBase {
  renderStyle() {
    return style;
  }
}

azakus:
Fixed with #290

PR #147, opened 2018-10-10
Implement icon-button in typescript

### LINKED ISSUE #95: Publish mwc-icon-button

azakus:
- icon-toggle is now icon-button, matching MDC
- If `offLabel` is not defined, button is always "on" and does not
toggle

Fixes #95


azakus:
This element is the first attempt at implementing #140

packages/icon-button/src/icon-button-base.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  updated() {
+    if (this.offIcon === '') {
+      this.on = true;
+    }
+  }
+
+  render() {
+    return html`
+      ${this.renderStyle()}
```
sorvell: Let's make sure there's an abstract `renderStyle`

packages/icon-button/src/icon-button-base.ts
```diff
@@ -0,0 +1,101 @@
...
+        }
+        this.dispatchEvent(new CustomEvent('MDCIconButtonToggle:change', {detail: evtData, bubbles: true}));
+      }
+    }
+  }
+
+  protected handleClick() {
+    if (this.offIcon !== '') {
+      this.on = !this.on;
+      this.mdcFoundation.handleClick();
```
sorvell: Are we sure that calling `handleClick` on the foundation and then `toggle` in the observer works? It seems like that could toggle the value 2x. If it doesn't, why not?

packages/icon-button/src/icon-button-base.ts
```diff
@@ -0,0 +1,101 @@
...
+  label = '';
+
+  @property({type: Boolean, reflect: true})
+  disabled = false;
+
+  @property({type: String})
+  icon = '';
+
+  @property({type: String})
+  offIcon = '';
```
sorvell: We discussed making the `offIcon` not optional, but if it is not, how is the behavior different? Would we throw if there's no `offIcon`? That would be strange.

packages/icon-button/src/icon-button-base.ts
azakus: This is already abstract in `BaseElement`

packages/icon-button/src/icon-button-base.ts
azakus: toggle just calls the adapter `addClass` twice, which uses `classList.add()`, and that should be a no-op to add the same class twice.

packages/icon-button/src/icon-button-base.ts
azakus: If there is no `offIcon`, the state never toggles and we do not emit an event.

PR #154, opened 2018-10-16
convert textfield to typescript

kr05:
This is my first attempt at converting the textfield component into typescript. A few things to note:

- I wasn't able to get the ripple to function properly. It kept giving me nasty "can't access classList of undefined" errors. I removed it, although I'm not sure what side effects it might have, if any. I also tried taking advantage of the ripple directive, but I couldn't get that to work either.

- I changed the "!this.fullWidth && this.label" check in the <label> element to simply "!this.fullWidth", as I was struggling to get the focused outline to work whenever there was no label.

- For some reason, the :focus styling was not working with the box variant of the textfield. I simply added the styling in the scss file, which a small change from :focus to :focus-within

- I was unable to access the initialization method, which is very handily provided by the component-element class. I got around this by simply calling a private initFactories method inside the createAdapter method, right before the adapter is created and returned. This ensures that the necessary factories are initialized before the adapter gets created, since it relies on these factories to do work.

Hope this is useful in some way, and I'm excited to be able to contribute in any way I can.


kr05:
I signed it!

elf-pavlik:
I find `value` getter and setter from #92 quite helpful, this PR seems to use `value` property differently.

jtbrne:
It doesn't seems like anything is happening with this phenomenal contribution. Is there any chance someone could tell us what you want or need in order to merge it and publish to npm?

packages/textfield/src/mwc-textfield.ts
```diff
@@ -0,0 +1,293 @@
...
+      'mdc-text-field--disabled': this.disabled,
+      'mdc-text-field--fullwidth': this.fullWidth,
+    };
+    return html`
+      ${this.renderStyle()}
+      <div class="mdc-text-field mdc-text-field--upgraded ${classMap(classes)}">
+        ${!this.fullWidth && this.icon ? html`<i class="material-icons mdc-text-field__icon" tabindex="0">${this.icon}</i>` : ''}
+        ${this._renderInput(this.value, this.required, this.type, this.placeHolder, this.label)}
+        ${!this.fullWidth ? html`<label class="mdc-floating-label ${this.value ? 'mdc-floating-label--float-above' : ''}" for="text-field">${this.label}</label>` : ''}
+        ${!this.fullWidth && this.outlined ? html`<div class="mdc-notched-outline">
```
kasper-f: Nice work with the typescript version.
I think a small extension would be very handy when extending the textfield.
eg. with a password field with a button at the end of the field toggling content viability (type=input/password)
```suggestion
        <slot name="suffix" slot="suffix"></slot> 
        ${!this.fullWidth && this.outlined ? html`<div class="mdc-notched-outline">
```

packages/textfield/src/mwc-textfield.ts
```diff
@@ -0,0 +1,293 @@
...
+      'mdc-text-field--box': this.box,
+      'mdc-text-field--outlined': !this.fullWidth && this.outlined,
+      'mdc-text-field--disabled': this.disabled,
+      'mdc-text-field--fullwidth': this.fullWidth,
+    };
+    return html`
+      ${this.renderStyle()}
+      <div class="mdc-text-field mdc-text-field--upgraded ${classMap(classes)}">
+        ${!this.fullWidth && this.icon ? html`<i class="material-icons mdc-text-field__icon" tabindex="0">${this.icon}</i>` : ''}
+        ${this._renderInput(this.value, this.required, this.type, this.placeHolder, this.label)}
```
kasper-f: Add possibility to embed another component, eg. a button toggling content viability for a password field. See also suggestion with suffix below.
```suggestion
        <slot name="prefix" slot="prefix"></slot>
        ${this._renderInput(this.value, this.required, this.type, this.placeHolder, this.label)}
```

kr05:
Thanks to everyone that has commented and made suggestions! I wanted to give a quick update on how this textfield has served my needs. 

I'm actually using a modified version of this PR in my own projects to much success. Most, if not all, of the modifications have to do with making the styling more flexible. For example, a few tweaks to the css have allowed me to do the following:

```css
mwc-textfield {
     --mdc-theme-primary: var(--app-secondary-color);
     --mdc-theme-label-color: var(--app-dark-text-color);
     --mdc-theme-font-family: 'Open Sans', sans-serif;
     --mdc-theme-font-weight: 700;
     /* and many more */
}
```
This was an absolute necessity for my own projects, since I needed the textfields to conform to my app's styling. The actual functionality has not changed, which simply means that the web components team has done a fantastic job in creating a solid base from which to build. I'll dedicate a couple of hours this Saturday to integrate some of the suggestions and hopefully get this closer to actually being published to npm.

kasper-f:
@kr05 It sound good you will give it an extra look,  just opened a pull request @ your branch with the slots embedded in the textfield and another branch im my fork to demonstrate the usage of the slot. 

See https://github.com/kr05/material-components-web-components/pull/1 

When I build the project the styling in mwc-textfield.scss failes unless i change the import:
from `@import '@material/mwc-icon/mwc-icon.scss';` to
 `@import '@material/mwc-icon/src/_mwc-icon.scss';`

Regarding styling I too is building the mwc-* components self in order to style the components.
I think I will create a pull request to help creating your own theme  when the original repository is a bit more up to date as it will change  the *.scss in every component. I think about adding an import in every component to a global theme-variables.scss, this way you can change the variables in one file and generate a component set that fit your style needs.

packages/textfield/src/mwc-textfield.ts
kr05: I looked into the changes, and I love the flexibility this provides, especially because I recently had to make use of textfields with a trailing icon and I would have loved this level of control. 

My solution was simply to make use of the "native" mwc-textfield way: the label property. I did have to manually add a click handler on the ```<i>``` tag provided, remove the fullWidth check, and be content with using material icons, but it was good enough for my needs. You can see it here:

image

markup before:
```html
${!this.fullWidth && this.icon ? html `<i class="material-icons mdc-text-field__icon" tabindex="0">${this.icon}</i>` : ''}
```

markup after:
```html
${this.icon ? html `<i @click="${this._onIconClick}" class="material-icons mdc-text-field__icon" tabindex="0">${this.icon}</i>` : ''}
```

As for integrating these changes, I think its best if we wait until a maintainer can go over the PR and your suggestion. Hopefully we can get an answer soon and have a more clear idea of what they want for this component.

kr05:
How would the theming work? I've been trying to find a solution for this problem, but I haven't been too successful. Do you have some kind of WIP or proof of concept? That sounds like an incredible thing to have!

kr05:
@elf-pavlik Thank you for the suggestion. I tried integrating the set and get methods for the value property, but it was throwing an error due to a clash with the property decorator. After some playing around, I discovered that simply adding "reflect: true" option solved the problem and allowed me to completely remove the event listener.

jtbrne:
Bump. I'm growing concerned that this is an abandoned project. @kr05 you might want to throw in the dense attribute...maybe that's causing the delay :)

packages/textfield/src/mwc-textfield.ts
```diff
@@ -0,0 +1,288 @@
...
+      } else {
+        if (this.mdcRoot.classList.contains(WITH_LEADING_ICON)) {
+          this.leadingIcon_ = iconFactory(iconElements[0]);
+        } else {
+          this.trailingIcon_ = iconFactory(iconElements[0]);
+        }
+      }
+    }
+
+    //I couldn't get rippleFactory to work. It keeps giving me "unable to access classList of undefined" errors.
```
luissardon: Why you don't use the ripple directive?

kr05: I was unable to get it to work unfortunately

luissardon: First the import

    import { ripple } from '@material/mwc-ripple/ripple-directive.js';

Then the ripple directive

    <div class="mdc-text-field mdc-text-field--upgraded ${classMap(classes)}" .ripple="${!outlined ? ripple({ unbounded: false }) : undefined}">

Finally you just need to redirect the inputs events to the root element in order to make the ripple work

    <input ... @focus="${this.handleInteractiveEvent}" @blur="${this.handleInteractiveEvent}">

Using something like

    handleInteractiveEvent(evt: MouseEvent) {
        const event = new CustomEvent(evt.type);
        this.mdcRoot.dispatchEvent(event);
    }

luissardon: **RippleOptions** accepts an *interactionNode* but doesn't work as expected because of the lifecycle; the input doesn't exist yet at that point, that's why I would try this approach.

kr05: I added the changes and it seems to work as expected. I'm not quite sure what "expected" means in the context of this textfield though. It doesn't throw an error, but I see no discernible difference by having the ripple directive versus not having it. I compared Google's own demo [here](https://material-components.github.io/material-components-web-catalog/#/component/text-field), this textfield with the ripple directive, and then without. Do you mind explaining what I should be expecting?

elf-pavlik:
I consider textfield as priority for most applications using forms. I see that this PR now has many conflicts, I wouldn't blame author for not rebasing it on latest master since this contribution didn't get much attention from maintainers. I hope maintainers could comment if this PR has chance of landing or they plan to create alternative PR for textfield themselves.

kr05:
> I consider textfield as priority for most applications using forms. I see that this PR now has many conflicts, I wouldn't blame author for not rebasing it on latest master since this contribution didn't get much attention from maintainers. I hope maintainers could comment if this PR has chance of landing or they plan to create alternative PR for textfield themselves.

Thanks for the comment, it motivated me to come back and update this PR. If you have a chance, I would appreciate it if you went over the updated PR and commented on what needs to be changed.

elf-pavlik:
Nice!

Now it shows 112 files changed in 124 commits, I guess something didn't go well with rebase?

kr05:
I wasn't sure what the best way to update this PR. I created a new branch from master, added the textfield changes, and merged that into this original textfield-updates branch. Rebasing from the textfield-updates branch directly was giving me too many conflicts, unfortunately :( GitHub claims this PR can be merged automatically since there are no conflicts to resolve, but I'm guessing that's not good enough. What do you suggest?

elf-pavlik:
Very weird, here it shows properly difference between branches: 10 commits and 4 files
https://github.com/material-components/material-components-web-components/compare/master...kr05:textfield-updates

kr05:
Good to know, thanks. I've been using a custom textfield in my projects that approximates the Material Design guidelines, but there's nothing like using an official implementation from Google. I know they are incredibly busy (can't wait for drawer and top-bar to get published) but hopefully this PR gets some attention from the maintainers soon!

elf-pavlik:
I asked github support about the issue with different diffs for the same branches, they replied:

> We generate the diff for a PR at the time it's created. If updates are made to the base branch (master) after creation, the compare diff and the PR diff would differ.

> As a workaround, changing the base branch on the PR to something else then back to master should regenerate the diff.

PR #167, opened 2018-12-03
Update lit-element, lit-html & remove event-stream

43081j:
This removes event-stream due to the recent vuln/code injection craziness.

It also updates lit-html and lit-element to the latest versions.

Does lerna not use package locks?


package.json
```diff
@@ -57,7 +57,7 @@
     "mocha": "^5.0.0",
     "node-sass-import": "^2.0.1",
     "node-watch": "^0.5.8",
-    "npm-run-all": "^4.1.1",
+    "npm-run-all": "^4.1.5",
```
43081j: This removes event-stream

packages/tab-scroller/src/mwc-tab-scroller.ts
```diff
@@ -54,7 +54,6 @@ export class TabScroller extends BaseElement {
   @query('.mdc-tab-scroller__scroll-content')
   protected scrollContentElement!: HTMLElement;
 
-  @eventOptions({passive: true})
```
43081j: `EventListenerOptions` doesn't contain `passive` and hasn't for some time, so i'm unsure why this suddenly doesn't compile. Maybe inside lit-element, we should be using `AddEventListenerOptions` instead ([see here](https://github.com/Microsoft/TypeScript/blob/28f8fdaccd78de926722e115157da32a50b77f4a/lib/lib.dom.d.ts#L33)).

This line shouldn't be removed but passive isn't valid so maybe a PR to lit needs doing first?

43081j: will be fixed by polymer/lit-element#326

43081j:
@azakus i noticed you bumped the same versions in master. still want this?

your commits dont seem to have changed `classMap` to `class-map`, so maybe this is still useful? and the run-all update

kr05:
Totally agree, just started a new project and am waiting on this fix

jamesmoore255:
Seconded, this needs to be resolved.

43081j:
yeah looks like this PR was completely overlooked from day 1 unfortunately. all 3 fixes will go out under other later PRs :man_shrugging:

PR #174, opened 2018-12-18
Correct usage documentation in README. Closes #34

### LINKED ISSUE #34: There is no index.js
From README.md:

> Import the element definition into your HTML page:
> 
> ```<script type="module" src="@material/mwc-icon/index.js"></script>```

But there is no `index.js` file. Looks like it should be `mwc-icon.js`.

This applies to all packages.

peterblazejewicz:
- no public export in index.js, provide correct filename

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ When you're ready to use the Material Web Components in your web application:
 
   1. Import the element definition into your HTML page:
 
-      ```<script type="module" src="@material/mwc-icon/index.js"></script>```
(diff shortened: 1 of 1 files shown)


aomarks:
The package.json maps `main` to `mwc-icon.js` so we don't need an `index.js`: https://github.com/material-components/material-components-web-components/blob/e158dc5722d30699fa0340e35a81402e4961e771/packages/icon/package.json#L5

aomarks:
Oops, actually you're right, I misread that we had explicitly put the `index.js` there. Thanks!

PR #176, opened 2018-12-26
Upgrade npm-run-all, fix broken link to malware in event-stream 3.3.6

jhnsntmthy:
There was a malware detected in a version of npm module event-stream v 3.3.6 and was removed from NPM, so upgrading the npm-run-all version fixes this dependency issue and removes the malware


jhnsntmthy:
I signed it!

PR #179, opened 2019-01-08
Fix ripple directive

straversi:
A change to lit-html (https://github.com/Polymer/lit-html/pull/562) means that the directive function must now wrap the ripple directive factory.


straversi:
Work already done by #164

PR #182, opened 2019-01-09
Adding checkedChange event to switch

alexander-patton:
allows parent elements to listen for when the checked property changes


frankiefu:
Fixed by https://github.com/material-components/material-components-web-components/pull/258. Now you can listen for `change` event.

PR #187, opened 2019-01-15
Move styles to to the new static styles API

### LINKED ISSUE #186: Use lit-element's `static get styles()` API
With lit-element 2.0, there is a new opt-in styling API that removes the need to put style rendering into the `render()` call.

This API will be faster when [`ShadowRoot.adoptedStyleSheets`](https://wicg.github.io/construct-stylesheets/) is implemented.

Dabolus:
This PR replaces the `renderStyles` method on each component with the new lit-element 2.0 static styles API. `sass-template.tmpl` has also been updated to generate a `CSSResult` using the new `css` tag, so that it can be easily consumed by the new static `styles` property.

Fixes #186


Dabolus:
Note that I flagged it as a WIP even if all components have already been migrated because there are still some things to tweak. For example, styles work like a charm on each component demo, but it seems like styles are not applied to the demos' `index.html`. Trying to figure out why.

azakus:
Wow, thanks for the work on this!
I'm going to wait on merging until the next lit-element rc is out, as it will coerce an array out of `get styles()`

Dabolus:
Just saw the new commits on lit-element. Would you like me to change the various
```javascript
static get styles() {
  return [style];
}
```

With just `static styles = style;`?

azakus:
Thanks for your patience! I've tested this with 2.0.0-rc.5, and everything works great!

PR #199, opened 2019-01-18
Drawer: Implement focus trapping

### LINKED ISSUE #191: Implement focus-trapping for drawer
Copy or adapt focus trapping techniques from dialog

straversi:
Fixes #191.

Note: Does not add list navigation integration because mwc-list component is not implemented.

### DIFF
--- a/packages/drawer/src/mwc-drawer.ts
+++ b/packages/drawer/src/mwc-drawer.ts
@@ -20,10 +20,19 @@ import MDCModalDrawerFoundation from '@material/drawer/modal/foundation.js';
 import MDCDismissibleDrawerFoundation from '@material/drawer/dismissible/foundation.js';
 import {strings} from '@material/drawer/constants.js';
 import {style} from './mwc-drawer-css.js';
+import 'wicg-inert/dist/inert.js';
+import 'blocking-elements/blocking-elements.js';
 
 declare global {
   interface HTMLElementTagNameMap {
     'mwc-drawer': Drawer;
+  }
+
+  interface Document {
+    $blockingElements: {
+      push(HTMLElement): void;
+      remove(HTMLElement): Boolean;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/drawer/src/mwc-drawer.ts
```diff
@@ -77,13 +85,12 @@ export class Drawer extends BaseElement {
       // TODO(sorvell): Implement list focusing integration.
       focusActiveNavigationItem: () => {
       },
-      // TODO(sorvell): integrate focus trapping.
-      trapFocus: () => {},
-      releaseFocus: () => {},
+      trapFocus: () => {
+        document.$blockingElements.push(this);
```
straversi: Ideally, `document.$blockingElements.remove(this);` would be in `releaseFocus`. However, this doesn't work because material-components-web calls releaseFocus and restoreFocus out of order in the dismissible adapter. Opened https://github.com/material-components/material-components-web/issues/4268 as a result of this discovery.

packages/drawer/src/mwc-drawer.ts
```diff
@@ -59,10 +68,9 @@ export class Drawer extends BaseElement {
         this._previousFocus = (this.getRootNode() as any as DocumentOrShadowRoot).activeElement as HTMLElement|null;
       },
       restoreFocus: () => {
+        document.$blockingElements.remove(this);
         const previousFocus = this._previousFocus && this._previousFocus.focus;
-        // Note, casting to avoid cumbersome runtime check.
-        const activeElement = (this.getRootNode() as any as DocumentOrShadowRoot).activeElement;
-        if (activeElement && this.mdcRoot.contains(activeElement) && previousFocus) {
+        if (previousFocus) {
```
straversi: Removed `activeElement && this.mdcRoot.contains(activeElement)` from this condition, because at this point in the transition lifecycle, the drawer has already been closed. Determining whether the old activeElement was in the drawer or not is irrelevant.

PR #204, opened 2019-01-25
SASS -> Sass (http://sassnotsass.com/)

toastal:
I don't want to sign into Google to sign an agreement -- since it's merely a typographical correction.


aomarks:
Sorry, we can't accept even the smallest contributions unless you sign the CLA. Please re-open if you would consider signing the SLA.

PR #206, opened 2019-01-28
Ripple: update element after setting interaction node

### LINKED ISSUE #203: How to make ripple work?
Shouldn't `mwc-ripple` do something in this example: https://dune-playroom.glitch.me/   I am using v0.3.5. The `div`s I lifted from the ripple demo. The button icon I added myself.  (BTW, why does the ripple demo import `mwc-icon`?)

straversi:
fixes #203

### DIFF
--- a/packages/ripple/src/mwc-ripple.ts
+++ b/packages/ripple/src/mwc-ripple.ts
@@ -46,7 +46,7 @@ export class Ripple extends LitElement {
 
   connectedCallback() {
     this.interactionNode = this.parentNode as HTMLElement;
-    this.performUpdate();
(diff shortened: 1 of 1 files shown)


packages/ripple/src/mwc-ripple.ts
```diff
@@ -46,6 +46,7 @@ export class Ripple extends LitElement {
 
   connectedCallback() {
     this.interactionNode = this.parentNode as HTMLElement;
+    this.performUpdate();
```
azakus: Replace with `super.connectedCallback()`

PR #207, opened 2019-01-28
[Fix - Tab Bar Component]: Fix Transpiling to ES5

daniel-nagy:
Calling `super` in a getter doesn't transpile to ES5. https://github.com/Microsoft/TypeScript/issues/338#issuecomment-347382190

### DIFF
--- a/packages/tab-bar/src/mwc-tab-bar.ts
+++ b/packages/tab-bar/src/mwc-tab-bar.ts
@@ -80,6 +80,24 @@ export class TabBar extends BaseElement {
     this.mdcFoundation.handleKeyDown(e);
   }
 
+  constructor() {
+    super();
+
+    const updateComplete = this.updateComplete;
+
+    Object.defineProperty(this, 'updateComplete', {
+      get() {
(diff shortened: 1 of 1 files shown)


packages/tab-bar/src/mwc-tab-bar.ts
```diff
@@ -80,6 +80,24 @@ export class TabBar extends BaseElement {
     this.mdcFoundation.handleKeyDown(e);
   }
 
+  constructor() {
+    super();
+
+    const updateComplete = this.updateComplete;
```
rictic: This reference becomes stale after the first update doesn't it? The `this.updateComplete` getter will return different promises at different times so that it always points to a promise that resolves for the _next_ update

aomarks:
Thanks for the PR! This is indeed a problem. We're going to address this by adding a new `_getUpdateComplete` method to `LitElement` (https://github.com/Polymer/lit-element/pull/745), and using it here (https://github.com/material-components/material-components-web-components/pull/307).

Issue #185, opened 2019-01-14
mwc-base - duplicate export 'TemplateResult'


amitrokhin:
Looks like `export {TemplateResult};` is redundant, since `lit-element` exports `TemplateResult` by itself (https://github.com/Polymer/lit-element/pull/415). Otherwise `polymer build` fails with `Duplicate export 'TemplateResult'` error.

kr05:
Hope this gets patched soon!

PR #209, opened 2019-01-30
Minimize direct imports of lit-html and lit-element

azakus:
mwc-base exports everything needed by the majority of elements.

This will reduce maintenance on updating lit versions.


straversi:
fixes #185

PR #210, opened 2019-01-30
TabBar: RTL support

straversi:
Add RTL support to `mwc-tab-bar`.

How: Create a property for [dir](http://mdn.io/dir) on `mwc-tab-bar` and pass it to to mwc-tab-scroller. Specifically, this gives `.mdc-tab-scroller__scroll-content` in the `mwc-tab-scroller` a computed `direction` value, which allows RTL scrolling to work properly.

fixes: #194


straversi:
WIP: still need to propagate RTL support to individual `mwc-tab`s

straversi:
WIP: need to fix rtl mixin

PR #211, opened 2019-01-31
Update and change out dependencies

azakus:
- Replace webpack with rollup for karma
- Remove most of the unused npm scripts
- Fix karma node resolver usage
- Remove webpack test/unit/index.js file
- Sauce runner configs from Polymer


karma.conf.js
```diff
@@ -15,143 +15,101 @@ See the License for the specific language governing permissions and
...
 module.exports = function(config) {
   config.set({
     basePath: '',
-    frameworks: ['mocha'],
+    frameworks: ['mocha', 'chai'],
     files: [
       {pattern: 'node_modules/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js', watched: false},
-      'test/unit/index.js',
+      {pattern: 'node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js', watched: false},
+      {pattern: 'test/unit/mwc-{button,fab,radio,ripple}*.js', watched: false},
```
straversi: Why only disable file watching on `button,fab,radio,ripple`?

scripts/build/rollup-demos.js
```diff
@@ -50,8 +50,6 @@ const rollupOptions = {
       module: true,
       jsnext: true,
       main: true,
-      browser: true,
-      modulesOnly: true,
```
straversi: Why were these true originally? I'm having trouble figuring that out.

karma.conf.js
azakus: This is actually only _testing` button, fab, radio, and ripple

scripts/build/rollup-demos.js
azakus: Copy/paste error 🙈

PR #213, opened 2019-01-31
TabBar: RTL support

straversi:
Support TabBars/Tabs.

How: (1) detect and set the `dir` attribute on `mwc-tab`. (2) Override Sass declarations from mdc-tab to detect `dir` on the host element.

Fixes: #194

### DIFF
--- a/packages/tab/src/mwc-tab.scss
+++ b/packages/tab/src/mwc-tab.scss
@@ -37,3 +37,22 @@ limitations under the License.
 .mdc-tab::-moz-focus-inner {
   border: 0;
 }
+
+@mixin mdc-rtl($root-selector: null) {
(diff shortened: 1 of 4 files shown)


packages/tab/src/mwc-tab.scss
```diff
@@ -36,4 +36,23 @@ limitations under the License.
...
+      }
+    }
+  } @else {
+    :host([dir="rtl"]) & {
+      @content;
+    }
+  }
+}
+
+.mdc-tab:not(.mdc-tab--stacked) .mdc-tab__icon + .mdc-tab__text-label {
```
azakus: Is overriding `mdc-rtl` not sufficient for this rule?

Issue #221, opened 2019-02-07
mwc-drawer: focus trap stops working when using slot="appContent"


frankiefu:
If the app content is put in the light dom of the `mwc-drawer` via `slot="appContent"` (for example in [demo/drawer/modal.html](https://github.com/material-components/material-components-web-components/blob/e1953b23a5eba52783dc604eea911fa835aefb4a/demos/drawer/modal.html#L38)), when the drawer is opened one could still "tab focus" on a focusable element outside of the drawer.

PR #224, opened 2019-02-11
Drawer: Fix focus trap for new appContent slot

straversi:
The focus trap behavior regressed when the `appContent` slot was introduced (Let's get integration tests). Frankie is right in #221: When the mdcRoot is added to $blockingElements, only the siblings of the parents are made inert. Since the `aside`, `scrim`, and `appContent` are all siblings, we have to get away with directly setting `inert` on appContent. We retain the $blockingElements usage for cases in which there are further ancestors.

Fixes: #221

### DIFF
--- a/packages/drawer/src/mwc-drawer.ts
+++ b/packages/drawer/src/mwc-drawer.ts
@@ -33,6 +33,10 @@ declare global {
       remove(HTMLElement): Boolean;
     }
   }
+
+  interface HTMLElement {
+    inert: Boolean;
+  }
 }
 
 export interface DrawerFoundation extends Foundation {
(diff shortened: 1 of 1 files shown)


packages/drawer/src/mwc-drawer.ts
```diff
@@ -86,6 +94,7 @@ export class Drawer extends BaseElement {
       },
       trapFocus: () => {
         document.$blockingElements.push(this);
+        this.appContent.inert = true;
```
frankiefu: What about just making `aside.mdc-drawer` the blocking element? e.g.
```
document.$blockingElements.push(this.mdcRoot);
```
Does that work?

straversi: The scrim is a sibling of the aside, and has to be clickable unfortunately.

straversi: (And mdc guidelines require the `.mdc-drawer-app-content` be a sibling of the `.mdc-drawer` and `.mdc-drawer-scrim`)

frankiefu: I see. Thanks for the explanation.

PR #227, opened 2019-02-15
Disable pointer events when button is disabled

### LINKED ISSUE #148: disabling a button should prevent handling click/tap events
I remember in paper-button disabling it would prevent handing click/tap events on it, which comes very handy when binding computed property, for example `?disabled=${!formValidates}`. Current button implementation seems to only style it as disabled but still leaves it functionality enabled.

balloob:
Disable pointer events when `<mwc-button>` is disabled.

Approach copied from how Polymer implements similar functionality: https://github.com/PolymerElements/iron-behaviors/blob/master/iron-control-state.js#L76-L95

Fixes #148

### DIFF
--- a/packages/button/src/mwc-button.ts
+++ b/packages/button/src/mwc-button.ts
@@ -14,7 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import {LitElement, html, property, observer, customElement, classMap} from '@material/mwc-base/base-element';
+import {LitElement, html, property, customElement, classMap} from '@material/mwc-base/base-element';
 import {style} from './mwc-button-css.js';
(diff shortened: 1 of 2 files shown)


packages/button/src/mwc-button.ts
```diff
@@ -34,6 +34,9 @@ export class Button extends LitElement {
   @property({type: Boolean})
   dense = false;
 
+  @observer(function(this: LitElement, value: boolean) {
+    this.style.pointerEvents = value ? 'none' : '';
+  })
   @property({type: Boolean})
```
frankiefu: Instead of adding an observer let's make the property `reflect: true` and in `mwc-button.scss` add:
```css
:host([disabled]) {
  pointer-events: none;
}
```

balloob: Done

frankiefu:
@balloob Thanks!

Issue #197, opened 2019-01-15
Remove "extra row" from top-app-bar


azakus:
This is complicated to implement, and not supported by all MDC wrappers.

straversi:
Discussion in #228 highlights how extra row is utilized for adding tabs to top app bars, including in pwa-starter-kit.

jsilvermist:
My use case for this is for a secondary title with some dynamic effects to produce a similar effect to what was present in the first Polymer Starter Kit (a long time ago).

straversi:
After reviewing, we preempted MDC Web by implementing this, and we'd like to avoid that and wait for them to release features first. So this issue will remain valid. `extraRow` will return when it is released in MDC Web.

PR #228, opened 2019-02-19
TopAppBar: remove extra row

straversi:
Remove "extra row" functionality from top app bar.

fixes: #197

### DIFF
--- a/demos/top-app-bar.html
+++ b/demos/top-app-bar.html
@@ -52,23 +52,14 @@ limitations under the License.
 
     mwc-top-app-bar[type="prominent"][dense] + .top-app-bar-adjust,
     mwc-top-app-bar[type="prominentFixed"][dense] + .top-app-bar-adjust,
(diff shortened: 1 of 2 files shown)


zoofadoofa:
How are you going to insert tabs into top-app-bar if you remove this slot? If the issue is with styling in regards to prominent you only need to remove the 'mdc-top-app-bar__section' from the extra row and prominent will work nicely

frankiefu:
[pwa-starter-kit](https://pwa-starter-kit.appspot.com) also needs "extra row" (or something equivalent)  to insert tabs into top-app-bar.

straversi:
Seems like we should re-evaluate #197. I'll add concerns there.

straversi:
Reviewed this further, and we preempted MDC Web by implementing this feature. We'd like to avoid that, so we're going to remove the `extraRow` property for now and reintroduce when MDC Web releases it.

PR #236, opened 2019-03-01
add mdc-button__label class

### LINKED ISSUE #235: mdc-button__label class is recommended
According [`mdc-button`](https://github.com/material-components/material-components-web/tree/master/packages/mdc-button):

> the `mdc-button__label` class may become mandatory for all cases in the future, so it is recommended to always include it to be future-proof

navigalia:
Add `mdc-button__label` class. Fixes #235 .

### DIFF
--- a/packages/drawer/src/mwc-drawer.ts
+++ b/packages/drawer/src/mwc-drawer.ts
@@ -74,8 +74,6 @@ export class Drawer extends BaseElement {
         this._previousFocus = (this.getRootNode() as any as DocumentOrShadowRoot).activeElement as HTMLElement|null;
       },
       restoreFocus: () => {
-        document.$blockingElements.remove(this);
-        this.appContent.inert = false;
         const previousFocus = this._previousFocus && this._previousFocus.focus;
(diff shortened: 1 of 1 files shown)


packages/button/src/mwc-button.ts
```diff
@@ -64,7 +64,7 @@ export class Button extends LitElement {
           ?disabled="${this.disabled}"
           aria-label="${this.label || this.icon}">
         ${this.icon ? html`<span class="material-icons mdc-button__icon">${this.icon}</span>` : ''}
-        ${this.label}
+        ${html`<span class="mdc-button__label">${this.label}</span>`}
```
elf-pavlik: Why do you need to nest templates here? Wouldn't just this work the same
```
<span class="mdc-button__label">${this.label}</span>
```

navigalia: Thanks @elf-pavlik, fixed. You'll see that I was also wrong with the head branch, so I had to open a new pull request, sorry.

PR #238, opened 2019-03-02
add mdc-button__label class

### LINKED ISSUE #235: mdc-button__label class is recommended

navigalia:
Add mdc-button__label class. Fixes #235.


frankiefu:
Thanks!

PR #250, opened 2019-03-20
remove extra character

eskan:
remove extra character


frankiefu:
Thanks for the PR. The fix is already in another [PR](https://github.com/material-components/material-components-web-components/pull/249/files#diff-9b9ec19374ffab5b1357a6be8ab02ce9L75) and we are planning to merge into master very soon.

PR #253, opened 2019-03-21
Use "classMap" as the entire value of the "class" attribute

gmincof:
Reference:
https://lit-html.polymer-project.org/guide/template-reference#classmap


gmincof:
I signed it!

PR #258, opened 2019-04-01
Forward `change` events out of form elements

### LINKED ISSUE #63: Ensure change events are forwarded out of mwc elements
Currently mwc elements rely on native events bubbling out of the shadowRoot for some cases. However, the <input>'s `change` event is not `composed: true` and therefore cannot be seen outside the shadowRoot.

This applies to at least `mwc-textfield` and `mwc-switch` and likely a bunch of other elements.

straversi:
Summary: Listen for and re-dispatch `change` events in `FormElement`s. The differences between the old and new event are that the new event will be untrusted and will have `src-element` and `target` equal to the custom element.

Fixes #63


kr05:
Has this already landed on npm? My switches don't emit the change event. Currently using v0.5.0.

PR #270, opened 2019-04-16
Reuse the proper CSS Custom Property for ripple opacity in icon-button

azakus:


packages/icon-button/src/mwc-icon-button.scss
```diff
@@ -28,10 +28,10 @@ limitations under the License.
 }
 
 // hack for top-app-bar ripple coloring of icon-buttons
-.mdc-ripple-upgraded {
+.mdc-icon-button.mdc-ripple-upgraded {
+  --mdc-ripple-fg-opacity: var(--mdc-icon-button-ripple-opacity, 0.12);
   &:focus::before, &:focus::after {
     background-color: currentColor;
     background-color: var(--mdc-theme-on-primary, currentColor);
-    opacity: var(--mdc-icon-button-ripple-opacity, 0.12);
```
frankiefu: The ripple looks lighter now when the icon-button inside top-app-bar is focused. It looks like `.mdc-ripple-upgraded--background-focused::before` has `opacity: .12` instead of using the custom property. Something like this maybe:
```
  &.mdc-ripple-upgraded--background-focused::before {
    opacity: var(--mdc-icon-button-ripple-opacity, 0.12);
  }
```

azakus:
Fixed a better way in #358

PR #277, opened 2019-05-07
fix(mwc-tab): DOMException The result must not have attributes

### LINKED ISSUE #276: (bug) document.createElement('mwc-tab') throws Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes
`document.createElement('mwc-tab')` throws 

```js
Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes
```

the issue is that the constructor sets `id` which updates the attribute

https://github.com/material-components/material-components-web-components/blob/8d2ab311e8cd24f2e82e91f89976e61d846098ad/packages/tab/src/mwc-tab.ts#L102

changing the property name or delaying adding the `id` until after the element is created (firstUpdated, connectedCallback, whatever) prevents the error.

stephantabor:
closes #276

There was already a unit test coverage for the bug, but the tests were not enabled for the tabs file. This enables the tabs tests and fixes the bug.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -85,7 +85,7 @@ module.exports = function(config) {
     files: [
       {pattern: 'node_modules/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js', watched: false},
       {pattern: 'node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js', watched: false},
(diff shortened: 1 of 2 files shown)


stephantabor:
signed

stephantabor:
:thinking:

PR #279, opened 2019-05-10
[mwc-radio] init Selection Controller when attached

eskan:
Hi,
the selection controller was init in constructor which is too early when radio button is created in a lit map to get a proper RootElement.
Selection controller is now created/get in connectedCallback()

### DIFF
--- a/packages/radio/src/mwc-radio.ts
+++ b/packages/radio/src/mwc-radio.ts
@@ -65,14 +65,14 @@ export class Radio extends FormElement {
 
   constructor() {
     super();
-    // Selection Controller is only needed for native ShadowDOM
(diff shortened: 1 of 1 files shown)


aomarks:
Thanks for the PR! Your investigation was correct, though due to some refactorings in the meantime and I ended up creating a new PR (#553). The bug here is now fixed and was just released to NPM in `0.10.0.

WeftDigital:
I am still experiencing this issue under specific circumstances, which I have attempted to replicate here:
https://jsbin.com/nofomes/edit?html,console,output

You'll notice there is timer that updates the webcomponent's data after three seconds. Once that update has rendered, the selection controller no longer works as expected.

dfreedm:
@WeftDigital Can you file your example as a separate issue? Thanks!

PR #291, opened 2019-07-12
Material 3.0.0

e111077:
Update material deps from ^1.0.0 to 3.0.0 and fix breakages.

lmk if you don't like my impl on top-app-bar. Decided not to use ResizeObserver because TS didn't like it and it's not supported in Safari.

Also added stronger types to `type` property on top-app-bar


packages/top-app-bar/src/mwc-top-app-bar-base.ts
```diff
@@ -139,12 +121,49 @@ export class TopAppBarBase extends BaseElement {
...
+  protected handleResize = () => {
+    this.mdcFoundation.handleWindowResize();
+  }
+
+  protected handleNavigationClick = () => {
+    this.mdcFoundation.handleNavigationClick();
+  }
+
+  protected registerListeners() {
+    this._navIconSlot.addEventListener('click', this.handleNavigationClick);
```
azakus: Can the click and scroll listeners be in the template? If so, I think that's a cleaner place to do that.

packages/ripple/src/ripple-directive.ts
```diff
@@ -19,6 +19,7 @@ import MDCRippleFoundation from '@material/ripple/foundation.js';
 import {MDCRippleAdapter} from '@material/ripple/adapter.js';
 import {style} from './mwc-ripple-global-css.js';
 import * as util from '@material/ripple/util.js';
```
azakus: Looks like only `supportsCssVariables` is imported now. Can you clean that up?

e111077: done

packages/top-app-bar/src/mwc-top-app-bar-base.ts
e111077: I moved the click listener in. I had it here before because I was running into a dumb issue where the foundation was not defined yet, but all worked if I just moved that call out into another function. scrollTarget may change from window to an HTML element, so I decided to keep that there.

PR #305, opened 2019-07-19
Remove circular dependency from <mwc-radio> by merging selection-controller into base.

aomarks:
Remove circular dependency from <mwc-radio> by merging selection-controller into base.


jsilvermist:
Out of curiosity, is this being done for best practice? Or is it due to it causing an error with something?

aomarks:
> Out of curiosity, is this being done for best practice? Or is it due to it causing an error with something?

It's due to some internal Google infrastructure that prefers not to deal with circular dependencies. Normally this style would be completely fine for TypeScript and ES modules.

aomarks:
@rictic review request (I can't actually add you since you aren't a collaborator, but I sent an invite).

rictic:
As I recall, the internal infra (that I think is open sourced as the bazel TypeScript rules) does work fine with circular dependencies within an individual ts_library, it just can't have them between ts_libraries

PR #306, opened 2019-07-20
Fix some type annotations

aomarks:
Fix some type annotations


aomarks:
@rictic review request (I can't actually add you since you aren't a collaborator, but I sent an invite).

packages/linear-progress/src/mwc-linear-progress-base.ts
```diff
@@ -85,7 +85,11 @@ export class LinearProgressBase extends BaseElement {
       ...addHasRemoveClass(this.mdcRoot),
       getPrimaryBar: () => this.primaryBar,
       getBuffer: () => this.bufferElement,
-      setStyle: (el: HTMLElement, property: string, value: string) => el.style[property] = value,
+      setStyle: (el: HTMLElement, property: string, value: string) =>
```
rictic: can we update the type of property here, instead of casting at the use site?

aomarks: This needs to match an interface defined in material components web. We can update that too, though the one place that calls this interface function calls it with `-webkit` prefixed strings, so then we need a cast there anyway. Maybe we should augment `CSSStyleDeclaration`. I just added a TODO for now, in the interest of incremental progress. Also pointed out what those two excluded properties are.

rictic: +1, thanks for the comment

PR #313, opened 2019-07-26
Fix tachometer

### LINKED ISSUE #312: Tachometer RAF not actually waiting until render

e111077:
There was a problem where our tachometer implementation using RAF would not paint the elements onto the page / do style recalculation before measuring.

I instead switched from using the lit element `test-fixture` and instead simply use `lit-html`'s render method and checking the first child and awaiting `updateComplete` + forcing style recalc w/ `offsetWidth`.

If there is no first child with `updateComplete` we just simply await a RAF.

Also changed the test harness interface to take an object of options rather than params.

fixes #312.

### DIFF
--- a/test/benchmark/helpers.ts
+++ b/test/benchmark/helpers.ts
@@ -1,5 +1,5 @@
 import { customElement, html, LitElement, property } from 'lit-element';
-import { TemplateResult } from 'lit-html';
+import { TemplateResult, render } from 'lit-html';
 
 declare global {
   interface Window {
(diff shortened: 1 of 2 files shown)


e111077:
CC: @sorvell @frankiefu

e111077:
tested normal material checkbox vs material checkbox with extra divs:

image

compared lit-html vs old RAF implementations

image (1)

Old implementation showed that it may be 0-4% slower and new implementation shows that it is _consistently_ slower at4-12% slower.

PR #316, opened 2019-07-29
Floating text label

### LINKED ISSUE #294: implement mwc-floating-label

e111077:
Implement floating label and put it in textfield

Fixes #294 

In this PR:
* Implemented floating text label as a directive
* I have a perf commit where you can look at my benchmarks
* Made tachometer more-correct where it now checks the last element and tries to await its updateComplete


packages/floating-label/src/mwc-floating-label-directive.ts
```diff
@@ -0,0 +1,31 @@
+import { MDCFloatingLabelAdapter } from '@material/floating-label/adapter.js';
```
azakus: Add license

test/benchmark/textfield/test-basic.ts
```diff
@@ -1,7 +1,7 @@
-import '@material/mwc-textfield';
+import '@material/mwc-textfield/mwc-textfield';
```
azakus: Add license

e111077: will be done in separate PR

PR #319, opened 2019-07-30
Implement Character counter

e111077:
we were using a textfield internal component which I pulled out. Here are perf results where "bundle" is their component bundled in vs "test-basic" being the new directive implementation

image


packages/textfield/src/character-counter/mwc-character-counter-directive.ts
```diff
@@ -0,0 +1,27 @@
+import { MDCTextFieldCharacterCounterAdapter } from '@material/textfield/character-counter/adapter.js';
```
azakus: Add license header

PR #331, opened 2019-07-31
Replace decorated HTMLSlotElement property type for emitDecoratorMetadata compatibility

aomarks:
Replace decorated `HTMLSlotElement` property for compatibility with TypeScript's `emitDecoratorMetadata`. With decorator metadata, the constructor is emitted into the runtime, which will cause an `HTMLSlotElement is undefined` error in Edge and IE11.


packages/top-app-bar/src/mwc-top-app-bar-base.ts
```diff
@@ -36,8 +36,12 @@ export class TopAppBarBase extends BaseElement {
   @query('.mdc-top-app-bar')
   protected mdcRoot!: HTMLElement;
 
+  // _actionItemsSlot should have type HTMLSlotElement, but when TypeScript's
+  // emitDecoratorMetadata is enabled, the HTMLSlotElement constructor will
+  // be emitted into the runtime, which will cause an "HTMLSlotElement is
+  // undefined" error in browsers that don't define it (e.g. Edge and IE11).
```
rictic: Good catch.

I wonder how hard it would be to disable `emitDecoratorMetadata` in google3. Potentially would have decent size savings.

aomarks: > I wonder how hard it would be to disable `emitDecoratorMetadata` in google3. Potentially would have decent size savings.

Yep!

PR #333, opened 2019-07-31
introduce clang-format

43081j:
im having trouble locally running it for now

cc @aomarks


43081j:
```
diff --git a/packages/checkbox/src/mwc-checkbox-base.ts b/packages/checkbox/src/mwc-checkbox-base.ts
index 1ccd645..1d7573b 100644
--- a/packages/checkbox/src/mwc-checkbox-base.ts
+++ b/packages/checkbox/src/mwc-checkbox-base.ts
@@ -14,32 +14,19 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 */
-import {
-  html,
-  FormElement,
-  property,
-  query,
-  observer,
-  HTMLElementWithRipple,
-  addHasRemoveClass,
-  RippleSurface,
-} from '@material/mwc-base/form-element.js';
-import {ripple} from '@material/mwc-ripple/ripple-directive.js';
-import MDCCheckboxFoundation from '@material/checkbox/foundation.js';
 import {MDCCheckboxAdapter} from '@material/checkbox/adapter.js';
+import MDCCheckboxFoundation from '@material/checkbox/foundation.js';
+import {addHasRemoveClass, FormElement, html, HTMLElementWithRipple, observer, property, query, RippleSurface,} from '@material/mwc-base/form-element.js';
+import {ripple} from '@material/mwc-ripple/ripple-directive.js';
```

this is what im seeing

aomarks:
https://github.com/material-components/material-components-web-components/pull/278 was submitted with the other `.clang-format`, so I think we should expect this one to reformat everything back to how it was earlier.

43081j:
i've never seen that before.. leaving a trailing comma in a one-liner.

and its 150+ characters long, but the google column limit is 100 usually 🤔 

something is off and you saw different results IIRC

aomarks:
Will merge this and see if I can fix the formatting in another PR.

aomarks:
Looks like the issue with trailing commas was just that clang-format doesn't *remove* them if they're there, and they were there because of the earlier config that broke it up into multiple lines.

I think the line-limit doesn't need to apply to imports, so that's fine.

All fixed in https://github.com/material-components/material-components-web-components/pull/334 I think.

PR #334, opened 2019-07-31
clang-format with new config

aomarks:
Updated the script to include test files, and to make sure we don't try to format files in node_modules.

Also removes trailing commas from imports. It looks like clang-format preserves them if they are there, but they were only there because of the other format config we had in the earlier PR.

cc @43081j


aomarks:
Oh, looks like we're now checking eslint in CI but didn't actually fix the existing errors yet!

PR #340, opened 2019-07-31
Add and update licenses

e111077:
update all 2018 licenses to 2019 and add licenses to all tests


aomarks:
We should keep the oldest license year.

e111077:
redoing branch anyway

PR #345, opened 2019-08-03
Notched outline

e111077:
Changed notched outline into an element from divs, adapters, and foundations. Also some more changes to benchmarking harness, and styling options for textfield and mwc-notched-outline

## Usage:
```js
html`
<mwc-notched-outline
    .width=${widthOfNotch}
    .open=${shouldOpen}
    class="mdc-notched-outline">
  <label .foundation=${floatingLabel()} for="text-field">
    Floating label!
  </label>
</mwc-notched-outline>`
```

## Performance:
It seems to be non-linear when you add a shadow root. The textfield needs to await the render of the label / notch before it can calculate the proper width because the element upgrades as an empty shadow root which makes the label not visible, and then once the update is complete, it stamps the slot (though this has repeatedly proved to consistently (at any N) only slow down the implementation by 0-2%). This seems to be a weird thing lit-element does.

In the following, `bundled` is the element how it was with adapter and foundation bundled in, and `element-props` is this latest implementation.

### 10 renders:
image

### 100 renders:
image

### 500 renders:
image

### Execution time vs instances rendered
The following chart shows the number of instances created vs time (in ms) of element creation.

yellow is this implementation, blue is the previous implementation, green is the difference, and the two horizontal lines are at 500 and 1000ms.
image


packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -239,12 +230,13 @@ export abstract class TextFieldBase extends FormElement {
     };
   }
 
-  protected getLabelAdapterMethods() {
+  protected getLabelAdapterMethods(): MDCTextFieldLabelAdapter {
     return {
-      floatLabel: (shouldFloat: boolean) =>
+      floatLabel: async (shouldFloat: boolean) =>
```
azakus: Why is this async?

Issue #348, opened 2019-08-06
Properly document components


azakus:
The stub README's are not good enough anymore 😜

aomarks:
From @samuelli in https://github.com/material-components/material-components-web-components/issues/325 (I closed as dupe of this one):

> Examples of incorrectness:
> 
> `import {Switch} from "@material/mwc-tab-bar"`
> Examples of not very useful:
> 
> `<mwc-tab-bar>sentiment_very_satisfied</mwc-tab-bar> should show usage with tabs`
> Examples of undocumented quirks:
> 
> `Use of icon property does not work without importing mwc-icon separately.`
> Examples of missing docs:
> 
> Events - digging through typings/code is not particularly feasible. For example, even after looking at the relevant code, it's difficult to find the event name. See https://github.com/material-components/material-components-web-components/blob/master/packages/tab-bar/src/mwc-tab-bar-base.ts#L148 which is defined as a string constant

aomarks:
Checklist!

[ ] button
[ ] checkbox
[ ] drawer
[ ] fab
[ ] formfield
[ ] icon
[ ] icon_button
[ ] linear_progress
[ ] radio
[ ] ripple
[ ] slider
[ ] snackbar
[ ] switch
[ ] tab
[ ] top_app_bar

orneryd:
> Checklist!
> 
> * [x]  button
> * [ ]  checkbox
> * [ ]  drawer
> * [x]  fab
> * [ ]  formfield
> * [ ]  icon
> * [x]  icon-button
> * [x]  icon-button-toggle
> * [ ]  linear-progress
> * [x]  radio
> * [ ]  ripple
> * [ ]  slider
> * [x]  snackbar
> * [ ]  switch
> * [ ]  tab
> * [ ]  top-app-bar
> * [x]  textarea
> * [x]  textfield

I'm disappointed that these are not in material web components :stuck_out_tongue_winking_eye:

jsilvermist:
@orneryd FTFY ;) https://mcwc-checklist.glitch.me/

![MCWC-Checklist](https://user-images.githubusercontent.com/10246291/63556625-910b5400-c513-11e9-8043-31400bf07b37.png)

azakus:
I believe we've separated all the individual components into their own issues, so I think we can close this master issue.

PR #355, opened 2019-08-09
Add expanded README for <mwc-snackbar>

aomarks:
Be sure to click `... > View file` to review the rendered version.

Much of the API text came from https://material.io/develop/web/components/snackbars/, but with some customization and additions to match our API.

cc https://github.com/material-components/material-components-web-components/issues/348


jsilvermist:
Shouldn't isOpen be a boolean not a number?

aomarks:
> Shouldn't isOpen be a boolean not a number?

Yes, thanks! Fixed.

PR #358, opened 2019-08-10
Refactor icon-button to accept SVG or image icons

### LINKED ISSUE #156: slot icon-button icons
It seems like the icon options could be greatly expanded by using slots for the off and on icons. The default slot content would be the standard material icons, but a slot would allow the use of any icon or svg. Just thought I would make the suggestion because the same would be true for the textfield and other components.

azakus:
Refactor `<mwc-icon-button>` to have slots, allowing SVG and image icons.

Icon must have `slot="onIcon"`attribute for non-toggling buttons, and an additional Icon with `slot="offIcon"` for toggling buttons

Example:
```html
<mwc-icon-button>
  <img src="on.png" slot="onIcon">
  <img src="off.png" slot="offIcon">
</mwc-icon-button>
```

TODO for PR:
- [x] svg example
- [x] image example
- [x] fix tests

Fixes #156

### DIFF
--- a/packages/icon-button/src/icon-button-base.ts
+++ b/packages/icon-button/src/icon-button-base.ts
@@ -27,7 +27,11 @@ export class IconButtonBase extends BaseElement {
 
   @query('.mdc-icon-button') protected mdcRoot!: HTMLElement;
 
-  @query('slot[name="offIcon"]') protected offIconSlot!: HTMLSlotElement;
+  // offIconSlot should have type HTMLSlotElement, but when TypeScript's
+  // emitDecoratorMetadata is enabled, the HTMLSlotElement constructor will
+  // be emitted into the runtime, which will cause an "HTMLSlotElement is
+  // undefined" error in browsers that don't define it (e.g. Edge and IE11).
+  @query('slot[name="offIcon"]') protected offIconSlot!: HTMLElement;
 
   @property({type: String}) label = '';
 
@@ -73,8 +77,8 @@ export class IconButtonBase extends BaseElement {
   }
 
   protected calculateShouldToggle() {
-    this.shouldToggle =
-        !(this.offIcon === '' && !this.offIconSlot.assignedNodes().length);
+    this.shouldToggle = this.offIcon !== '' ||
+        (this.offIconSlot as HTMLSlotElement).assignedNodes().length > 0;
   }
 
   focus() {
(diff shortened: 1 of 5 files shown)


packages/icon-button/src/icon-button-base.ts
```diff
@@ -27,28 +27,36 @@ export class IconButtonBase extends BaseElement {
 
   @query('.mdc-icon-button') protected mdcRoot!: HTMLElement;
 
+  @query('slot[name="offIcon"]') protected offIconSlot!: HTMLSlotElement;
```
aomarks: There's a very annoying bug that means we can't use the `HTMLSlotElement` type in this position, explained here: https://github.com/material-components/material-components-web-components/blob/c8e187b2f74a62902d418b6b5963052e49e19fa0/packages/formfield/src/mwc-formfield-base.ts#L75

It means that when you compile with `emitDecoratorMetadata` (which we do internally), then this code breaks in IE/Edge.

packages/icon-button/src/icon-button-base.ts
```diff
@@ -58,18 +66,27 @@ export class IconButtonBase extends BaseElement {
...
-    if (this.offIcon !== '') {
+    if (this.shouldToggle) {
       this.on = !this.on;
       this.mdcFoundation.handleClick();
     }
   }
 
+  protected calculateShouldToggle() {
+    this.shouldToggle =
+        !(this.offIcon === '' && !this.offIconSlot.assignedNodes().length);
```
aomarks: Mild nitty preference for `this.offIcon !== '' || this.offIconSlot.assignedNodes().length > 0` -- but your call.

packages/icon-button/src/icon-button-base.ts
```diff
@@ -83,9 +100,14 @@ export class IconButtonBase extends BaseElement {
         aria-hidden="true"
         aria-label="${this.label}"
         ?disabled="${this.disabled}">
-        <i class="material-icons mdc-icon-button__icon">${this.offIcon}</i>
-        <i class="material-icons mdc-icon-button__icon mdc-icon-button__icon--on">${
-        this.icon}</i>
+        <span class="mdc-icon-button__icon">
+          <slot name="offIcon" @slotchange="${
+        this.calculateShouldToggle}"><i class="material-icons">${
```
aomarks: Maybe if you add a newline before the `<i>` then clang won't try and split on the expression

aomarks: Nice use of fallback content!

packages/icon-button/src/icon-button-base.ts
azakus: Done

packages/icon-button/src/icon-button-base.ts
azakus: Done

packages/icon-button/src/icon-button-base.ts
azakus: Yep, that works!

aomarks:
Add CHANGELOG entry? (I've been adding them to the top-level one, not sure about the status of the per-package ones).

azakus:
> Add CHANGELOG entry? (I've been adding them to the top-level one, not sure about the status of the per-package ones).

I added a line to the top-level changelog.

PR #360, opened 2019-08-12
Add lit-analyzer to linting

### LINKED ISSUE #359: Integrate lit-analyzer in CI
It's essentially a command line version of @runem's ts-lit-plugin so it does type checking of templates, ensures that the html is well formed, that `@property` declarations are correct, the works. It's configurable, and generally super awesome. 

We've got it integrated into all internal google lit-html and lit-element projects that are built with TypeScript. I haven't tried out the CLI much yet (internally we integrate via bazel) but from what I've seen I expect it to be high quality.

azakus:
Set up linting in Travis as well

Fixes #359


e111077:
Can you please split up bootstrap so it only sets up the workspace and does not build or lint?

PR #361, opened 2019-08-13
README for mwc-fab

aomarks:
Be sure to click `... > View file` to see the rendered version.

Also updates all demos to load the correct Roboto font widths, and updates the FAB demo to match the examples in the README.

https://github.com/material-components/material-components-web-components/issues/348


aomarks:
> Can you npmignore the images?

Ok, I added an `.npmignore` to every package (as recommended here https://github.com/lerna/lerna/issues/296#issuecomment-261080345).

In a follow up, I'm planning to write a `prepack` script that replaces all the local image references with github URLs, so that these READMEs render correctly on npmjs.com even without the images being part of the tarball.

PR #363, opened 2019-08-13
Script to rewrite README images as permanent URLs

aomarks:
This way we don't need to publish images in our NPM tarballs (they're already in our `.npmignore` files), but the READMEs will still render with images on npmjs.com, webcomponents.org, etc.

It resolves master to a SHA, and uses that in a raw.githubusercontent.com URL, so that if we change or move images files around, any given version of the README published to NPM won't be
affected.

I made this a prepack script, so that it runs automatically before we publish. (We would not commit those changes, since on GitHub it makes more sense to just the relative path like always).

For example, here's what it did to the FAB README:

static-image-example

Also threw in a small fix to our demo CSS Roboto font weights (my mistake from earlier PR).

### DIFF
--- a/scripts/static-urlify-readme-images.ts
+++ b/scripts/static-urlify-readme-images.ts
@@ -1,0 +1,81 @@
+/**
+ * Copyright 2019 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


e111077:
This means we write and commit them as relative but only prepack will publish them as direct?

aomarks:
> This means we write and commit them as relative but only prepack will publish them as direct?

Yep!

scripts/static-urlify-readme-images.ts
```diff
@@ -0,0 +1,81 @@
...
+import * as path from 'path';
+import {URL} from 'url';
+
+const mwcRepoRoot = path.resolve(__dirname, '..', '..');
+
+const githubRaw =
+    `https://raw.githubusercontent.com/material-components/material-components-web-components`;
+
+// Matches markdown image syntax like `![description](url "tooltip")`
+const imageRegexp = /(\!\[.*?\]\()([^) ]+)(.*?\))/g;
```
e111077: why is `?` required in the last capture group?

aomarks: It makes it non-greedy, otherwise it would match `descriptionany subsequent characters as long as there's eventually another )`

azakus: Could also be `([^)]*\))`, but either is fine with me

PR #366, opened 2019-08-14
Better readme for button

azakus:
Related #348


stramel:
Wow, nice job @azakus! 👍

azakus:
@stramel Thanks. We're going to try to do this format for all the components.

PR #373, opened 2019-08-15
Fix broken checked property in <mwc-radio>

aomarks:
Previously, we didn't trigger the `SelectionController` (the class that synchronizes radio button groups separated by a shadow root) when the checked property was set, so the previously checked radio in a group was not unchecked when a new one was checked.

This was more complicated than just triggering it from the existing property `@observer`, because that gives us batched async changes (via the `UpdatingElement` updated callback), so we can't reconstruct the order of checked sets correctly. Replaces the observer with a setter that can handle this synchronously.

### DIFF
--- a/packages/radio/src/mwc-radio-base.ts
+++ b/packages/radio/src/mwc-radio-base.ts
@@ -24,11 +24,46 @@ export class RadioBase extends FormElement {
 
   @query('input') protected formElement!: HTMLInputElement;
 
-  @property({type: Boolean})
-  @observer(function(this: RadioBase, checked: boolean) {
-    this.formElement.checked = checked;
-  })
-  checked = false;
+  private _checked = false;
+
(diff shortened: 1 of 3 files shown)


packages/radio/src/mwc-radio-base.ts
```diff
@@ -24,11 +24,46 @@ export class RadioBase extends FormElement {
 
   @query('input') protected formElement!: HTMLInputElement;
 
-  @property({type: Boolean})
-  @observer(function(this: RadioBase, checked: boolean) {
-    this.formElement.checked = checked;
-  })
-  checked = false;
+  private _checked = false;
```
azakus: Should this be `protected`?

aomarks: Discussed offline, private makes sense for now.

PR #374, opened 2019-08-16
fix(infrastructure): use postinstall instead of custom script

### LINKED ISSUE #381: Use postinstall instead of custom script
If we declare postinstall, npm executes it after npm install automatically.
(https://docs.npmjs.com/misc/scripts)
So what about to using postinstall instead of custom script?
Thank you :)

TroyTae:
closes #381


.travis.yml
```diff
@@ -7,11 +7,11 @@ addons:
   chrome: stable
 
 before_script:
-  - npm run bootstrap
+  - npm install
```
aomarks: I think we can just remove this line now. Travis will automatically run `npm ci` (since we have a `package-lock.json`, otherwise it would automatically run `npm install`), so we want to avoid installing twice (https://docs.travis-ci.com/user/languages/javascript-with-nodejs/). I confirmed that `postinstall` is run during `npm ci` also.

aomarks:
Thanks for the PR @TroyTae -- it does seem convenient to roll the bootstrap step into install! Just one comment, then LGTM.

.travis.yml
TroyTae: Oh I don't know about this thank you :)
I reflected it!

PR #375, opened 2019-08-16
Add eslint-plugin-lit

stramel:
Adds `eslint-plugin-lit` to help with linting lit-html. 

https://github.com/43081j/eslint-plugin-lit

/cc @43081j

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -4,6 +4,7 @@ extends:
   - plugin:@typescript-eslint/eslint-recommended
   - plugin:@typescript-eslint/recommended
   - plugin:wc/recommended
+  - plugin:lit/all
 globals:
   goog: false
 env:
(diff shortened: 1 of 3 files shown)


aomarks:
How much does this overlap with https://github.com/runem/lit-analyzer#readme which we integrated in https://github.com/material-components/material-components-web-components/pull/360?

43081j:
a lot for you. you probably get most of what you need from the analyzer.

its an odd situation now. people with TS can get almost everything from lit-analyzer, but those without (or those that use ESLint at all) benefit from having the lint plugin.

the analyzer can do type checks (i.e. check property bindings are passing the right type). the rest is covered by the lint plugin/is equal.

stramel:
As far as I can tell, I don't see any overlap?

lit-analyzer: https://github.com/runem/lit-analyzer/blob/master/docs/readme/rules.md
eslint-plugin-lit: https://github.com/43081j/eslint-plugin-lit#supported-rules

PR #393, opened 2019-08-21
Fix Travis Config

azakus:


.travis.yml
```diff
@@ -7,10 +7,8 @@ addons:
   chrome: stable
 
 before_script:
-  - >-
-  npm run format && git diff --exit-code || (echo -e '\n\033[31mERROR:\033[0m
-  Project is not formatted. Please run "npm run format".' && false)
-
+  - npm run build
```
aomarks: Why not put all of this in `script`? I never quite understood why I would put something in `before_script` vs `script`. I think maybe it's useful when you have a matrix of builds, where there's a common `script`, but each platform has a different `before_script`.

azakus: yeah, I can do that

azakus:
We again have working builds! And they fail!

PR #394, opened 2019-08-21
Textfield validation

### LINKED ISSUE #54: How would you add "step" as a property to the element?
Hi,

Thanks for the awesome elements.
I would like to inherit from [mwc-text-field](https://github.com/material-components/material-components-web-components/blob/master/packages/textfield/mwc-textfield.js) having the possibility to set `step` (used with the `number` type).

The first idea that springs to mind is by redefining this function:

     _renderInput({value, required, type, placeHolder, label}) {
        return html`<input type="text" type$="${type}" placeholder$="${placeHolder}" required?="${required}" class$="mdc-text-field__input ${value ? 'mdc-text-field--upgraded' : ''}" id="text-field" value="${value}" aria-label$="${label}">`;
      }

IS this the right path?
The problem is that the template simply calls:

     ${this._renderInput({value, required, type, placeHolder, label})}

So, if I want to keep this function pure, I don't have access to the passed value. I would effectively have to add it to the `properties`, the `constructor` and `_render`. While adding it to the constructor would be dead easy, the same cannot be said about properties and _render.

So... what's a good way to extend `mwc-text-field` so that it accepts a new property that wasn't there before?

(Note: this question can be applied to ANY element...)

Merc.

### LINKED ISSUE #57: Initial error message for TextField component
How to render a text field in an invalid state with a custom error message?

e111077:
Add validation to `mwc-textfield` and `mwc-textarea`.

Fixes #54.
Fixes #57.

I want to write tests, but I want to figure out our testing infra first.


aomarks:
> I want to write tests, but I want to figure out our testing infra first.

I wrote a test. It was fine after I figured out how to edit the karma config to isolate the build pipeline to one file at a time. LMK if I can help. We should definitely improve the infrastructure to make this really fast an easy, but for now editing the config is ok.

e111077:
rebased over #402

e111077:
added tests

PR #396, opened 2019-08-21
Fall back to msMatchesSelector in findAssignedElement util

aomarks:


aomarks:
> LGTM if you also add a direct dependency on `@material/dom`

Done, in all 3 packages where we use it.

PR #414, opened 2019-08-23
Testing

### LINKED ISSUE #398: Make tests green
Travis was disabled for a week because of a badly formated `.travis.yml` file, and we need to get tests back to green.

### LINKED ISSUE #415: Move tests to TS

e111077:
## What is in here
* Tests now in TS
* Tests now all pass in evergreens but still failing in IE and Safari 9
* Using Koa Karma proxy server and auto-amd-ifies files
* Can now run individual tests using `npm run test -- --packages mwc-icon-button*,mwc-button`
* Can now debug tests using `npm run test:debug -- --packages mwc-textfield`
* Made some "canonical" tests for textfield
	* Made some small fixes to textfield's focus
* Disabled tests for ie and safari 9
  * Failed on all counts due to compilation
  * Will file bug unless you want me to keep them on
* Travis still fails when running tests locally (will have to look into that)

Fixes #415.
Fixes #398.


aomarks:
Wow! This looks like a fantastic improvement, will review tomorrow.

README.md
```diff
@@ -78,9 +78,12 @@ Below are instructions for setting up project development.
 1. Run a development server with `npm run dev`
     - View the demos by accessing `<dev server url>`/demos/index.html
     - This will also build the project.
-1. Run tests with `npm run test`
+1. Run tests with `npm run build:tests && npm run test`
 1. Run benchmarks with `npm run build:tests && npm run test:bench`
 
+*Note:* You can debug specific tests by running `npm run test:debug -- --packages mwc-button,mwc-fab`
+or you may use wildcards such as `npm run test:debug -- --packages mwc-icon-button*`
```
aomarks: Maybe also add that you can run on just one browser with e.g. `--browsers=ChromeHeadless` (which is what I usually do when iterating on a test, since it's the fastest).

package.json
```diff
@@ -60,6 +63,12 @@
     "karma-rollup-preprocessor": "^6.1.2",
     "karma-sauce-launcher": "^2.0.2",
     "karma-sourcemap-loader": "^0.3.7",
+    "koa": "^2.8.1",
+    "koa-esm-transform": "^1.0.0-pre.1",
+    "koa-karma-proxy": "^1.0.0-pre.4",
+    "koa-mount": "^4.0.0",
+    "koa-node-resolve": "^1.0.0-pre.6",
+    "koa-static": "^5.0.0",
```
aomarks: Can we remove `rollup` and friends from dependencies now? And maybe the karma babel plugin?

packages/icon-button-toggle/src/icon-button-toggle-base.ts
```diff
@@ -55,6 +55,13 @@ export class IconButtonToggleBase extends BaseElement {
   }
 
   protected handleClick() {
+    const onIconSlotted = this.querySelector('[slot="onIcon"]');
+    const offIconSlotted = this.querySelector('[slot="offIcon"]');
+
+    if (!(this.onIcon || onIconSlotted) || !(this.offIcon || offIconSlotted)) {
```
aomarks: Why is this check needed? Foundation blows up or something? Maybe add a comment?

packages/icon-button-toggle/src/icon-button-toggle-base.ts
```diff
@@ -55,6 +55,13 @@ export class IconButtonToggleBase extends BaseElement {
   }
 
   protected handleClick() {
+    const onIconSlotted = this.querySelector('[slot="onIcon"]');
```
aomarks: Shouldn't this be `this.shadowRoot.querySelector`?

test/src/unit/mwc-textfield.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    fixt.remove();
+  }
+});
+
+const isUiInvalid =
+    (element: TextField) => {
+      return !!element.shadowRoot!.querySelector(`.${cssClasses.INVALID}`);
+    }
+
+suite('mwc-textfield - validation:');
```
aomarks: Interesting, I haven't seen this flat style before. I've always seen tests nested within the suite callback, like it is in the mocha docs (https://mochajs.org/#tdd). I guess the advantage of doing it in this flat style is that there's less code indentation? But I think I find it harder to visually parse the code this way, because the suites boundaries aren't as obvious now.

packages/icon-button-toggle/src/icon-button-toggle-base.ts
e111077: There were tests that were written that checked for this behavior. It seemed like it was missing. Thanks for pointing it out because I wanted to know if you thought if we should remove those tests or add the functionality back in?

test/src/unit/mwc-textfield.test.ts
e111077: This is qunit style ([expand below](https://github.com/material-components/material-components-web-components/pull/414/files/207eb83fd7163c4434672d919e908b6aadd9313d#diff-a2a3b7b0c9c3b4b93b4aebf4e3ec3cfbR108)). It doesn't allow for nesting. I'd kinda prefer the nesting tbh, but all the tests were already written this way.

@azakus @sorvell @frankiefu: any reason as to why y'all chose `qunit` over `TDD`?

packages/icon-button-toggle/src/icon-button-toggle-base.ts
e111077: this is purposefully seeking the light dom's projected elements (not the default icon). Thought it would be cleaner than selecting the slot and using `assignedNodes`. (I also wonder what the performance indications are for `assignedNodes`)

test/src/unit/mwc-textfield.test.ts
e111077: Here are the mocha docs for qunit https://mochajs.org/#qunit

e111077: affected parties are okay to switching to TDD

packages/icon-button-toggle/src/icon-button-toggle-base.ts
e111077: I think I'm going to revert this behavior

e111077:
Ready for final review

e111077:
Rebasing...

e111077:
rebased

azakus:
I think we can consider this "fixing" #398.

PR #417, opened 2019-08-23
Re-implement `centerTitle` for MWC Top App Bar

### LINKED ISSUE #413: Add back TopAppBar centerTitle
While not explicity in the spec, it _does_ seem useful, and there is _an_ example of it in the Material example apps:

azakus:
- Updated CHANGELOG
- Added screenshots
- Fixed `<mwc-top-app-bar centerTitle dense prominent>` styling

Fixes #413


packages/top-app-bar/src/mwc-top-app-bar-base-base.ts
```diff
@@ -42,6 +42,8 @@ export abstract class TopAppBarBaseBase extends BaseElement {
 
   private _scrollTarget!: HTMLElement|Window;
 
+  @property({type: Boolean, reflect: true}) centerTitle = false;
```
e111077: Does this need to be reflected? What are the reasons for?

packages/top-app-bar/src/mwc-top-app-bar.scss
```diff
@@ -50,3 +50,22 @@ limitations under the License.
     transition: padding 150ms $mdc-animation-standard-curve-timing-function;
   }
 }
+
+// add support for centered title
+%centerPadding {
```
e111077: I'm actually okay with this usage of sass

packages/top-app-bar/src/mwc-top-app-bar-base-base.ts
azakus: It was `reflect: true` previously, but I don't think that actually matters. I'll remove that.

PR #420, opened 2019-08-24
actually enable sauce launcher

azakus:
😡


e111077:
Can we prevent one of these travis runs from doing sauce tests? They seem to be sharing resources and conflicting with each other

PR #448, opened 2019-08-31
Add mdc-fab__label wrapper

abraham:
This matches the [MDC documented example](https://github.com/material-components/material-components-web/blob/master/packages/mdc-fab/README.md).

```html
<button class="mdc-fab mdc-fab--extended">
  <span class="material-icons mdc-fab__icon">add</span>
  <span class="mdc-fab__label">Create</span>
</button>
```


abraham:
@azakus ping.

Issue #454, opened 2019-09-04
internalize textfield


e111077:
fixes Polymer/internal#298.

PR #458, opened 2019-09-05
Changes to Textfield, line-ripple, and floating-label to build internally

e111077:
- **INTERNAL BREAKING** Changed interface to:
  - LineRipple: foundation -> lineRippleFoundation
  - CharCounter: foundation -> characterCounterFoundation
  - FloatingLabel: foundation -> floatingLabelFoundation
- **INTERNAL BREAKING** added following properties to global `Element` interface
  - lineRippleFoundation?: MDCLineRippleFoundation
  - characterCounterFoundation?: MDCTextFieldCharacterCounterFoundation
  - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
- Renamed textfield's directive applications
- Removed implicit any's in textfield caused by ValidityState

related #454

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - **BREAKING** Changed interface to:
   - LineRipple: foundation -> lineRippleFoundation
-  - CharCounter: foundation -> characterCounterFoundation
+  - CharCounter: foundation -> charCounterFoundation
   - FloatingLabel: foundation -> floatingLabelFoundation
 - **BREAKING** added following properties to global `Element` interface
   - lineRippleFoundation?: MDCLineRippleFoundation
-  - characterCounterFoundation?: MDCTextFieldCharacterCounterFoundation
+  - charCounterFoundation?: MDCTextFieldCharacterCounterFoundation
   - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
 - internal fixes to textfield's directives and removed implicit any's
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!--## Unreleased-->
+## Unreleased
+
+- **BREAKING** Changed interface to:
+  - LineRipple: foundation -> lineRippleFoundation
+  - CharCounter: foundation -> characterCounterFoundation
+  - FloatingLabel: foundation -> floatingLabelFoundation
+- **BREAKING** added following properties to global `Element` interface
+  - lineRippleFoundation?: MDCLineRippleFoundation
+  - characterCounterFoundation?: MDCTextFieldCharacterCounterFoundation
+  - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
+- internal fixes to textfield's directives and removed implicit any's
 
 ## [0.8.0] - 2019-09-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
-- **BREAKING** Changed interface to:
-  - LineRipple: foundation -> lineRippleFoundation
-  - CharCounter: foundation -> charCounterFoundation
-  - FloatingLabel: foundation -> floatingLabelFoundation
-- **BREAKING** added following properties to global `Element` interface
-  - lineRippleFoundation?: MDCLineRippleFoundation
-  - charCounterFoundation?: MDCTextFieldCharacterCounterFoundation
-  - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
-- internal fixes to textfield's directives and removed implicit any's
+- textfield
+  - **BREAKING** added following properties to global `Element` interface
+    - lineRippleFoundation?: MDCLineRippleFoundation
+    - charCounterFoundation?: MDCTextFieldCharacterCounterFoundation
+    - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
+  - compiles with more strict flags
 
 ## [0.8.0] - 2019-09-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 - textfield
-  - **BREAKING** added following properties to global `Element` interface
-    - lineRippleFoundation?: MDCLineRippleFoundation
-    - charCounterFoundation?: MDCTextFieldCharacterCounterFoundation
-    - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
   - compiles with more strict flags
 
 ## [0.8.0] - 2019-09-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## Unreleased
-
-- textfield
-  - compiles with more strict flags
+<!-- ## Unreleased -->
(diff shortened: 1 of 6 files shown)


packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -16,25 +16,42 @@ limitations under the License.
...
+  interface Element {
+    foundation?: MDCLineRippleFoundation|MDCTextFieldCharacterCounterFoundation;
+  }
+}
+
+type ValidityStateProp =
+    'badInput'|'customError'|'patternMismatch'|'rangeOverflow'|'rangeUnderflow'|
+    'stepMismatch'|'tooLong'|'tooShort'|'typeMismatch'|'valid'|'valueMissing';
+type CustomValidityState = {
+  [prop in ValidityStateProp]: boolean;
```
aomarks: Can you do this?

```ts
interface CustomValidityState {
  badInput: boolean;
  customError: boolean;
  ...
}
```

Seems easier to read.

packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -16,25 +16,42 @@ limitations under the License.
...
 import {characterCounter, CharacterCounter} from './character-counter/mwc-character-counter-directive.js';
 
+// must be done to get past lit-analyzer checks
+declare global {
+  interface Element {
+    foundation?: MDCLineRippleFoundation|MDCTextFieldCharacterCounterFoundation;
+  }
+}
+
+type ValidityStateProp =
```
aomarks: Mind adding a comment explaining why we need this? And if it's something to do with `readonly` properties, then maybe you can use an `Exclude` similar to this (but not sure I understand this case): https://github.com/material-components/material-components-web-components/blob/a87438e7f9c736f80b6596d05887f5bfcba69d3a/packages/linear-progress/src/mwc-linear-progress-base.ts#L92

packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -16,25 +16,42 @@ limitations under the License.
...
+import {MDCTextFieldCharacterCounterFoundation} from '@material/textfield/character-counter/foundation.js';
 import MDCTextFieldFoundation from '@material/textfield/foundation.js';
 import {ifDefined} from 'lit-html/directives/if-defined.js';
 
 import {characterCounter, CharacterCounter} from './character-counter/mwc-character-counter-directive.js';
 
+// must be done to get past lit-analyzer checks
+declare global {
+  interface Element {
+    foundation?: MDCLineRippleFoundation|MDCTextFieldCharacterCounterFoundation;
```
aomarks: (not blocking this PR, just discussion) There must be a better way to apply directives to an element that doesn't require inventing a fake property. @justinfagnani thoughts on that?

aomarks: Can we use more specific names so that we don't get collisions in the types? E.g. the next time another package has to augment element with `foundation`, they won't be able to.

packages/textfield/src/mwc-textfield-base.ts
e111077: all of the params of `ValidityState` are readonly. Users are unable to create one, only read it via `HTMLInputElementInstance.validity`. I'm trying to merge the custom validity partial with it and return it as type `ValidityState` for type checking which is why I can't use the keys of `ValidityState` and need my own.

aomarks:
Also can you update the PR description so that it's clear which element this fixes? And really, it's fixing typescript and lit-analyzer issues that in theory aren't internal specific (and now I wonder why we didn't catch them earlier in GitHub CI?).

e111077:
it seems as if build is more strict internally

packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -16,25 +16,67 @@ limitations under the License.
...
+// must be done to get past lit-analyzer checks
+declare global {
+  interface Element {
+    floatingLabelFoundation?: MDCFloatingLabelFoundation;
+    lineRippleFoundation?: MDCLineRippleFoundation;
+    charCounterFoundation?: MDCTextFieldCharacterCounterFoundation;
+  }
+}
+
+interface CustomValidityState {
```
aomarks: I just found out that in TypeScript 2.8 you can now use a mapped type to remove `readonly`, so I think you could use:

```ts
{ -readonly [P in keyof ValidityState]: ValidityState[P] };
```

https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#improved-control-over-mapped-type-modifiers

e111077: this is much more bueno

packages/textfield/package.json
```diff
@@ -22,6 +22,10 @@
     "lit-html": "^1.0.0",
     "tslib": "^1.10.0"
   },
+  "devDependencies": {
```
aomarks: I think these have to be dependencies, or else when a user compiles the textfield typings and don't also depend on these by some other means, then they'll get an unknown type error (since the textfield typings reference these typings)

CHANGELOG.md
```diff
@@ -4,7 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!--## Unreleased-->
+## Unreleased
+
+- **BREAKING** Changed interface to:
```
aomarks: Are any of these actually breaking changes? None of these properties are part of the public API, I would think? So maybe nothing to put in the changelog actually (except maybe that it compiles with more strict flags, though we still don't know which ones).

e111077: well the global `Element` part would be public breaking

aomarks: For which use case?

CHANGELOG.md
```diff
@@ -6,15 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
-  - lineRippleFoundation?: MDCLineRippleFoundation
-  - charCounterFoundation?: MDCTextFieldCharacterCounterFoundation
-  - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
-- internal fixes to textfield's directives and removed implicit any's
+- textfield
+  - **BREAKING** added following properties to global `Element` interface
+    - lineRippleFoundation?: MDCLineRippleFoundation
+    - charCounterFoundation?: MDCTextFieldCharacterCounterFoundation
+    - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
+  - compiles with more strict flags
```
aomarks: I don't think we should bother mentioning the flags unless it's specific, it's just kinda noise otherwise.

CHANGELOG.md
```diff
@@ -6,15 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
-  - LineRipple: foundation -> lineRippleFoundation
-  - CharCounter: foundation -> charCounterFoundation
-  - FloatingLabel: foundation -> floatingLabelFoundation
-- **BREAKING** added following properties to global `Element` interface
-  - lineRippleFoundation?: MDCLineRippleFoundation
-  - charCounterFoundation?: MDCTextFieldCharacterCounterFoundation
-  - floatingLabelFoundationtion?: MDCFloatingLabelFoundation
-- internal fixes to textfield's directives and removed implicit any's
+- textfield
+  - **BREAKING** added following properties to global `Element` interface
```
aomarks: I still don't really think this is a breaking change. Nobody other than us should care about these properties. We should eventually try to get rid of the gross global type augmentation (maybe file an issue), but this doesn't seem like anything a user needs to know about.

packages/textfield/src/mwc-textfield-base.ts
justinfagnani: There isn't currently. This is a pretty common need though...

justinfagnani: Is that why all the `foundation` properties are renamed in this PR?

e111077:
tests are success!

packages/textfield/src/mwc-textfield-base.ts
e111077: correct!

aomarks: https://github.com/material-components/material-components-web-components/blob/d6f763fb51d20ee669f4aeef850611506ff42b94/packages/textfield/src/mwc-textfield-base.ts#L35

PR #463, opened 2019-09-09
Fix focus button textfield

e111077:
- mwc-button
  - Added custom focus and blur functions that cause ripple
- mwc-textfield
  - Focus function now focuses the native input and activates the caret
  - Blur appropriately blurs the input

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- mwc-button
+  - Added custom focus and blur functions that cause ripple
+- mwc-textfield
+  - Focus function now focuses the native input and activates the caret
+  - Blur appropriately blurs the input
 
 ## [0.8.0] - 2019-09-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
-- mwc-button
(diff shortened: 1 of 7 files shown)


e111077:
canigetanamen?

CHANGELOG.md
```diff
@@ -4,7 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- mwc-button
```
aomarks: Mind matching the format of what's below? I was doing Changed/Added/Fixed sections. Also a little more detail would be nice. Why would a user care about new focus and blur functions?

packages/base/package.json
```diff
@@ -15,6 +15,9 @@
     "lit-html": "^1.0.0",
     "tslib": "^1.10.0"
   },
+  "devDependencies": {
+    "@material/ripple": "^3.0.0"
```
aomarks: This must be a dependency, because `MDCRippleFoundation` is the base class of an exported class. If a user imported it, and didn't happen to also depend on the ripple package, then TyepScript wouldn't be able to resolve that import.

packages/button/src/mwc-button-base.ts
```diff
@@ -34,10 +35,34 @@ export class ButtonBase extends LitElement {
...
+
   protected createRenderRoot() {
     return this.attachShadow({mode: 'open', delegatesFocus: true});
   }
 
+  focus() {
+    const buttonElement = this.buttonElement;
+    if (buttonElement) {
+      const ripple = buttonElement.ripple;
+      if (ripple) {
```
aomarks: Add a comment? Why does ripple need to know about focus? I imagine it just as background animation that you wouldn't think has any bearing on what's focused.

PR #473, opened 2019-09-12
Fixed MWC Tab Bar to set MDC Tab height using CSS vars.

aomarks:
Fixed MWC Tab Bar to set MDC Tab height using CSS vars.


azakus:
Mind doing the same thing with padding for #483?

azakus:
Looks like some upstream issue is causing the sass compiler to fail?

aomarks:
Oh sorry, this PR is replaced by https://github.com/material-components/material-components-web-components/pull/482 which is already merged. This is an upstream of an internal change from the MDC web team, and it depends on changes from MDC web 4.0.0, which is why #482 merges into the temporary `canary` branch. Closing.

PR #479, opened 2019-09-16
add select() to text fields

### LINKED ISSUE #68: mwc-textfield / area should expose selectionStart / selectionEnd
I've just hit a use-case where I needed to know the selectionStart/end of the mwc-textfield. Right now this is not being exposed. I have fixed the issue with my local fork via a getter/setter, like so:

```
  	get selectionStart() {
		return this._input.selectionStart;
  	}

  	set selectionStart(selection) {
		this._input.selectionStart = selection;
  	}

	set selectionEnd(selection) {
		this._input.selectionStart = selection;
  	}

	get selectionEnd() {
		return this._input.selectionEnd;
  	}
```

I'm happy to do a PR if the team is OK with using getter/setters for selection positions.

### LINKED ISSUE #461: mwc-textfield & mwc-textarea should implement select()
**I'm submitting a:**
- [ ] bug report
- [x ] feature request

**What OS are you using?**
macOS Mojave, 10.14.6

all (not bug, but needed in all browsers)

**Current behavior:**
There is currently no way to select text content without reaching into element's shadow root via `textElem.shadowRoot.activeElement.select()`

**Expected behavior:**
`mwc-textfield` should implement `HTMLInputElement.select()`, and the properties `selectionStart` and `selectionEnd`. (The same applies to `mwc-textarea`| `HTMLTextAreaElement`.)

**Steps to reproduce:**
see below

**Related code:**

```js
import { LitElement, html, css } from 'lit-element';
import '@material/mwc-textfield';
import '@material/mwc-textarea';

class MyElement extends LitElement {
  static get properties() {
    return {
      lorem: {
        type: String,
      }
    }
  }

  constructor() {
    super();
    this.lorem = `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.`;
  }

  _selectAll(evt) {
    evt.target.shadowRoot.activeElement.select();
  }

  static get styles() {
    return [
      css`
        :host {
          display: flex;
          flex-direction: column;
        }
        .mwctext {
          margin: 0.5em;
        }
      }`
    ]
  }

  render(){
    return html`
      <mwc-textfield 
        class="mwctext"
        label="username" 
        name="username"
        value="joeslick"
        @focus="${this._selectAll}">
      </mwc-textfield>
      <mwc-textarea
        class="mwctext"
        label="lorem" 
        name="lorem"
        value="${this.lorem}"
        @focus="${this._selectAll}">
      </mwc-textarea>
    `;
  }
}

customElements.define('my-element', MyElement);
```

**Other information:**
[Example on Stackblitz](https://stackblitz.com/edit/mwc-text-elements)

43081j:
pretty simple one

fixes #461.
fixes #68.

LMK if you're up for merging it. also if you think we need to dispatch an event or anything alongside calling `select()` (as thats what focus/blur methods do).

cc @azakus @e111077

### DIFF
--- a/test/src/unit/mwc-textfield.test.ts
+++ b/test/src/unit/mwc-textfield.test.ts
@@ -203,4 +203,27 @@ suite('mwc-textfield:', () => {
       }
     });
   });
+
+  suite('select', () => {
+    let element: TextField;
+
+    setup(async () => {
+      fixt = await fixture(basic);
+
+      element = fixt.root.querySelector('mwc-textfield')!;
+    });
+
+    test('selects the input text', () => {
+      const input = element.shadowRoot!.querySelector('input');
+
+      element.select();
+      assert.equal(element.shadowRoot!.activeElement, input);
+    });
+
+    teardown(() => {
+      if (fixt) {
+        fixt.remove();
+      }
+    });
+  });
 });
--- a/test/src/unit/mwc-textfield.test.ts
+++ b/test/src/unit/mwc-textfield.test.ts
@@ -221,7 +221,7 @@ suite('mwc-textfield:', () => {
       element.select();
 
       assert.equal(input.selectionStart, 0);
-      assert.equal(input.selectionEnd, 5);
+      assert.equal(input.selectionEnd, 6);
     });
 
     teardown(() => {
--- a/test/src/unit/mwc-textfield.test.ts
+++ b/test/src/unit/mwc-textfield.test.ts
@@ -214,10 +214,14 @@ suite('mwc-textfield:', () => {
     });
 
     test('selects the input text', () => {
-      const input = element.shadowRoot!.querySelector('input');
+      const input = element.shadowRoot!.querySelector('input')!;
+
+      input.value = 'foobar';
 
       element.select();
-      assert.equal(element.shadowRoot!.activeElement, input);
+
+      assert.equal(input.selectionStart, 0);
+      assert.equal(input.selectionEnd, 5);
     });
 
     teardown(() => {
--- a/test/src/unit/mwc-textfield.test.ts
+++ b/test/src/unit/mwc-textfield.test.ts
@@ -240,10 +240,12 @@ suite('mwc-textfield:', () => {
       element = fixt.root.querySelector('mwc-textfield')!;
     });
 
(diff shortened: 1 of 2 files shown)


e111077:
While you're here, do you mind implementing [`setSelectionRange`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange)?

test/src/unit/mwc-textfield.test.ts
```diff
@@ -203,4 +203,27 @@ suite('mwc-textfield:', () => {
...
+      fixt = await fixture(basic);
+
+      element = fixt.root.querySelector('mwc-textfield')!;
+    });
+
+    test('selects the input text', () => {
+      const input = element.shadowRoot!.querySelector('input');
+
+      element.select();
+      assert.equal(element.shadowRoot!.activeElement, input);
```
e111077: MDN says that not all browsers focus the selection. Perhaps use `getSelection`?

43081j: seems there's some funkiness with `getSelection` on the window/root. it doesn't work with inputs well if at all (googled a bit and seemed to say you should just access the input's start/end ranges directly).

43081j: i could maybe just `focus` before the select, here in the test?

e111077: That wouldn't necessarily test select but focus. I think your test for setSelectionRange would also work here

e111077:
awaiting tests and internal tests

43081j:
two tests failed, one is me being off-by-one. the other is a bit weird, got a 0 rather than 4 index on the setSelectionRange. will have a poke around

e111077:
fixes #68

PR #481, opened 2019-09-16
Start using MDC Web canary versions

aomarks:
MDC Web has a number of breaking changes coming up for the 4.0.0 release (changelog: https://github.com/material-components/material-components-web/blob/v4.0.0-canary.e851d4f40.0/CHANGELOG.md#breaking-changes). These are found in their `develop` branch, and are automatically published to NPM with the `canary` label. This is also the version of the code that runs inside Google.

The MWC repo currently depends on the released 3.0.0 packages. We do not yet have any branch which is compatible with these upcoming 4.0.0 changes. The code running inside Google is the same as our `master` branch, which is why we have some bugs there (such as ripple not working).

Ultimately we likely need a more complex staged release process that matches MDC Web, but I think for now it's going to be easiest to just actually depend on these pinned canary versions of MDC Web, *and actually release MWC to NPM with those dependencies*. That way GitHub, NPM, and Google will all be in sync.

So I've created a new `canary` branch where we'll add compatibility for 4.0.0. Once all the packages are working, we can merge it to `master`. This PR starts us off with:

- Bump every MDC web dependency in our packages to the latest MDC web canary.

- Add a script which finds the latest MDC web canary version, and bumps all the MDC web deps. Run with `npm run bump-all-mdc-web-deps`. We'll want to run this regularly, and before we publish. It uses the `=` semver operator because otherwise we might pick up more breaking changes before we are ready for them.

- Temporarily add some dependencies to the new `touch-target` packages, because those are missing in the latest MDC Web canary package.jsons (this is already fixed, so we can remove this the next time MDC Web canary is published).

- A temporary compilation fix to `mwc-slider`, just to get things building. I don't understand the new API yet.

- Add a new `mdc-button__ripple` element to our `mwc-button` template, which fixes ripple there.


aomarks:
> preapproval. Really waiting for internal screenshot tests

Note this PR is merging into the canary branch, not master. I wasn't planning on testing this in Google immediately, I wanted to sync up to MDC Web 4.0.0 incrementally (and hopefully split up the work). Once the canary branch looks good, we can merge it to master.

PR #488, opened 2019-09-18
Fix outlined textfield notch when initially rendered hidden

### LINKED ISSUE #444: mwc-textfield label rendering incorrectly when value is preset.

**I'm submitting a:**

- [ X] bug report
- [ ] feature request

**What OS are you using?**
Windows 10

All (Chrome 76)

**Current behavior:**
Label doesn't render correctly when value is preset.

image

```
<mwc-textfield       
          label="Name"
          value="test"
          outlined
          required
          validationMessage="Name is required"
        >
```

Label also doesn't render correctly when present with binding.  

`value="${this.newSlideShowName || ''}"`

image

e111077:
fixes #444 by providing the `mwc-textfied.layout()` method.

In this PR:
* added `layout` function
* notch performs layout when label is changed
* fixed hidden bug where foundation was being recreated too often
  * caused bug where changing `maxLength` caused foundation to unset invalid
  * had to pull logic out of foundation and into `registerValidationAttributeHandler`

The issue: when the label is floating and the element is outlined, but the label has no width, e.g.
`mwc-textfield[outlined][label="something"][value="something"][hidden]`

the mdc dataflow will calculate the size of the notch as zero. When hidden is removed, the notch will not update in size. Here we give them the `layout` method.

It is not possible to know when the textfield is changed from `display: none` to visible without setting a resize observer on the label. I wanted to avoid that because of possible render trashing and the overall heaviness of `resizeObserver` and a polyfill. Instead, it is up to the user to call layout when they change the visibility of the element.


packages/textfield/README.md
```diff
@@ -195,6 +195,7 @@ type ValidityTransform = (value: string, nativeValidity: ValidityState) => Parti
 | `checkValidity() => boolean`   | Returns `true` if the textfield passes validity checks. Returns `false` and fires an [`invalid`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event) event on the textfield otherwise.
 | `reportValidity() => boolean`   | Runs `checkValidity()` method, and if it returns false, then ir reports to the user that the input is invalid.
 | `setCustomValidity(message:string) => void`   | Sets a custom validity message (also overwrites `validationMessage`). If this message is not the empty string, then the element is suffering froma  custom validity error and does not validate.
+| `layout() => Promise<void>`   | Used to recalculate the layout of the notched outline when the label is floated. e.g. The outlined element has `display: none`, the label is floating, and then changed to `display: block`. You must call `layout`.
```
aomarks: This documentation is a little unclear to me, seems like not quite complete sentences. Can you rephrase it so that it would be very clear for a user if and when they would need to call this method?

packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -221,26 +229,25 @@ export abstract class TextFieldBase extends FormElement {
...
 
   updated(changedProperties: PropertyValues) {
-    const charCounter =
-        changedProperties.get('charCounter') as boolean | undefined;
-
-    // update foundation only when charCounter goes from false to true
-    if (!charCounter && this.charCounter) {
+    const maxLength = changedProperties.get('maxLength') as number | undefined;
+
+    const mLChangedFromUndef = maxLength === -1 && this.maxLength !== -1;
```
aomarks: the `mLC` prefix looks kinda weird. How about `maxLengthBecameDefined` and `maxLengthBecameUndefined`?

packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -221,26 +229,25 @@ export abstract class TextFieldBase extends FormElement {
...
+    const mLChangedToUndef =
+        maxLength !== undefined && maxLength !== -1 && this.maxLength === -1;
+    // update foundation only when maxLength goes from undefined (-1) to true
```
aomarks: Can you replace this with a comment explaining *why* we update (or is it re-create?) the foundation in this case (rather than describing *what* the code does)?

packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -486,11 +514,29 @@ export abstract class TextFieldBase extends FormElement {
...
+    if (this.label) {
+      await this.layout();
+    }
+  }
+
+  async layout() {
+    await this.updateComplete;
+
+    if (this.labelElement && this.outlineElement) {
+      // sometimes notchOutline in adapter is called before label is rendered.
```
aomarks: Can you expand the explanation of why we need this method (similar to what's in the PR description would be good: "If X is initially display:none or otherwise has the wrong width, then the foundation will calculate the wrong size and do its initial layout incorrectly, etc."

packages/textfield/src/mwc-textfield.scss
```diff
@@ -28,6 +28,10 @@ limitations under the License.
   @extend %material-icons;
 }
 
+.hidden, .hidden::before, .hidden::after {
+  display: none;
```
aomarks: Would `visibility:hidden` be better? Would it actually avoid this whole issue, because `visibility:hidden` shouldn't affect the computed size?

e111077: In this case, the display has to be none or else if you don't set a helper or a char counter, you'll have excess whitespace under the textfield. Also the problem is when they set `<mwc-textfield>` to display none, not what we do internally.

packages/textfield/src/mwc-textfield-base.ts
e111077: :+1: I was initially trying to not use the term "undefined" because of the overload

packages/textfield/src/mwc-textfield.scss
e111077: Though, we do need the elements to actually exist in the DOM or else we have to deal with recreating the char counter foundation or the foundation needing the char counter div when it is not rendered, and working around this would complicate the js logic more than this styling.

aomarks: > In this case, the display has to be none or else if you don't set a helper or a char counter, you'll have excess whitespace under the textfield.

Not sure I understand, why would `visibility:hidden` cause extra whitespace? Doesn't `hidden` mean layout is normal but the browser just doesn't draw anything?

> Also the problem is when they set `<mwc-textfield>` to display none, not what we do internally.

Yeah, makes sense, though if our `.hidden` rule could use `display:hidden`, then we could recommend users use the `hidden` attribute instead of setting `display:none` in order to avoid this problem.

packages/textfield/README.md
```diff
@@ -195,6 +195,7 @@ type ValidityTransform = (value: string, nativeValidity: ValidityState) => Parti
 | `checkValidity() => boolean`   | Returns `true` if the textfield passes validity checks. Returns `false` and fires an [`invalid`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event) event on the textfield otherwise.
 | `reportValidity() => boolean`   | Runs `checkValidity()` method, and if it returns false, then ir reports to the user that the input is invalid.
 | `setCustomValidity(message:string) => void`   | Sets a custom validity message (also overwrites `validationMessage`). If this message is not the empty string, then the element is suffering froma  custom validity error and does not validate.
+| `layout() => Promise<void>`   | Used to recalculate the layout of the notched outline when the label is floated. This would be used in situations when an outlined textfield has a floating label, but the outline notch is too small. Calling `layout` will resize the outline notch to fit the label.
```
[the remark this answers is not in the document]
e111077: That is correct, it is when the user sets the element to display none. There is no way to know this without using a resize observer, but I don't want to have to use polyfills or support that for IE

packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -486,11 +520,34 @@ export abstract class TextFieldBase extends FormElement {
...
+    }
+  }
+
+  async layout() {
+    await this.updateComplete;
+
+    if (this.labelElement && this.outlineElement) {
+      /* When the textfield automatically notches due to a value and label
+       * being defined, the textfield may be set to `display: none` by the user.
+       * this means that the notch is of size 0px. We provice this function so
```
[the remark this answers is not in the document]
e111077: d is just a squished cl

packages/textfield/src/mwc-textfield.scss
e111077: For the first part: the styles in `MDCTextfield` add specific height styling to the `.mdc-text-field-helper-line` if it is simply present. If we set `visibility:hidden`, then the underbar still has that height since `visibility` does not change the height unlike `display:none`.

For the second one, `visibility:hidden` does not fit all the use cases of `display: none`. What if they don't want it to be visible and for it to not take up space on the page.

packages/textfield/README.md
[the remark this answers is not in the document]
aomarks: Is this reply for the other comment?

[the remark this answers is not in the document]
e111077: yes it is. Trying a new extension in vscode. It is not going well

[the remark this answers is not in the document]
e111077: That is correct, it is when the user sets the element to display none. There is no way to know this without using a resize observer, but I don't want to have to use polyfills or support that for IE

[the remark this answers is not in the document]
e111077: okay for real now it's updated

PR #490, opened 2019-09-19
update textfield value on input instead of change

### LINKED ISSUE #474: No way to get textfield's value in an `@input` event
**I'm submitting a:**

- [ ] bug report
- [x] feature request

**Current behavior:**

I can attach a `@input` event to `TextField`
```html
<mwc-textfield @input="${e => (No way to get textfield value here)}">
```

Doing a `e.target.value` above will give blank as the value is only updated on the `change` event internally in the Component.

**Expected behavior:**
```html
<mwc-textfield @input="${e => (Should have a way to get value here)}">
```

e111077:
fixes #474

Can't test these changes with our current setup.

Benchmark results came up with creation being 0-3% faster for reasons unknown. On typing couldn't really be tested because we would need selenium or puppeteer to test. Changing values via JS showed inconclusive on change (though eyeballing a bit slower maybe?).

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "lint:imports": "node scripts/check-imports.js",
     "lint": "npm run lint:typescript && npm run lint:lit",
     "lint:typescript": "eslint \"packages/**/*.ts\"",
-    "lint:lit": "lit-analyzer \"packages/**/*.ts\" --strict --quiet",
+    "lint:lit": "lit-analyzer \"packages/**/*.ts\" --strict",
     "test": "npm run build:tests && karma-proxy start --single-run",
     "test:debug": "npm run test -- --single-run false --browsers Chrome",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "lint:imports": "node scripts/check-imports.js",
     "lint": "npm run lint:typescript && npm run lint:lit",
     "lint:typescript": "eslint \"packages/**/*.ts\"",
-    "lint:lit": "lit-analyzer \"packages/**/*.ts\" --strict --quiet",
+    "lint:lit": "lit-analyzer \"packages/**/*.ts\" --strict",
```
azakus: Why remove the `--quiet` argument?

e111077: because it's not quiet internally and fails on build. This will not fail build externally if there is a problem with the linter.

azakus: Ugh, `--quiet` not setting the exit code is not what I expected 😕

Issue #471, opened 2019-09-12
textarea filled disabled floating label too high


e111077:
floating label animation in `textarea[disabled][value="non empty string"]` is set to the same animation as notched outline in `textarea[outlined]`.

This styles is maintained by mwc not mdc.

![image](https://user-images.githubusercontent.com/5981958/64823243-c41d9200-d56b-11e9-89cc-7f2bb3ecbc90.png)

Issue #470, opened 2019-09-12
textarea outlined disabled has a filled background


e111077:
`textarea[outlined][disabled]` seems to show a filled background. This style is maintained by mwc not mdc.

![image](https://user-images.githubusercontent.com/5981958/64823190-a2bca600-d56b-11e9-9b25-e7da5d31eb6e.png)

PR #492, opened 2019-09-19
fix disabled styles on textarea

e111077:
before:

image

after:

image

~awaiting confirmation that this is correct from design~


e111077:
approved by design

e111077:
fixes #470

e111077:
fixes #471

PR #498, opened 2019-09-24
Move unit tests into package directories

aomarks:
- Move unit test files into individual package directories. We agreed we preferred this stylistically, and its also more compatible with how files are laid out internally in Google, which makes it easier to import and run there.
- Selecting specific tests from the `npm test` command still works.
- This means eslint and lit-plugin is now running on test TypeScript files, so all issues found by that are now fixed.
- Found some additional issues from internal testing:
  - Fixed implicit `any` errors with test elements.
  - Replaced casts to `any` with some new `Internals` interface casts.
  - Replaced Chai `isEmpty` and `isOk` calls with equivalents since we are on Chai 2.x internally and those functions don't exist at that version.
  - Replaced all `expect` calls with `asserts`, for consistency, and so we don't need to import the `expect` function internally.
  - Add missing `@license` annotations to copyright headers.


test/tsconfig.json
```diff
@@ -4,13 +4,14 @@
     "composite": true,
     "outDir": "./lib",
     "tsBuildInfoFile": ".tsbuildinfo",
-    "rootDir": "./src",
+    "rootDir": "../",
```
e111077: is this file still necessary here, can it be removed or put somewhere else?

aomarks: Seems like this is still a reasonable place for it, since it's the tsconfig for tests. It could be at the top level, but this seems fine too?

PR #504, opened 2019-09-27
Update mwc-radio to MDC 4

### LINKED ISSUE #485: Update mwc-radio for MDC 4
Part of https://github.com/material-components/material-components-web-components/issues/480

mwc-radio needs an additional node for ripple in MDC 4

e111077:
fixes #485.

### DIFF
--- a/packages/radio/src/mwc-radio-base.ts
+++ b/packages/radio/src/mwc-radio-base.ts
@@ -135,7 +135,7 @@ export class RadioBase extends FormElement {
 
   protected render() {
     return html`
-      <div class="mdc-radio">
+      <div class="mdc-radio" .ripple=${ripple()}>
         <input
           class="mdc-radio__native-control"
           type="radio"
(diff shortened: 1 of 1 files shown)


packages/radio/src/mwc-radio-base.ts
```diff
@@ -147,6 +149,7 @@ export class RadioBase extends FormElement {
           <div class="mdc-radio__outer-circle"></div>
           <div class="mdc-radio__inner-circle"></div>
         </div>
+        <div class="mdc-radio__ripple" .ripple=${ripple()}></div>
```
aomarks: I guess we should do the same for `<mwc-button>`, where I created the new ripple node but didn't move the directive. It seemed to be working anyway, not sure I understand this directive though.

azakus: I think the styling for the ripple is dependent only on the class names that the directive will create, but we probably should be mirroring the MDC template layout for maximum compatibility.

aomarks: I just tried moving the `<mwc-button>` `ripple` directive from the button to the ripple node, and the ripple stopped working. Instead, the whole button uniformly changes color on click, but there is no ripple effect.

e111077: this ripple thing is hella confusing let me tell you

e111077: Looking at their demo's code. I am wrong. Though it works, the ripple directive needs to stay on the root

e111077: Can confirm, the ripple still works

PR #505, opened 2019-09-27
update mwc-slider to MDC 4

### LINKED ISSUE #486: Update mwc-slider for MDC Web 4
Part of https://github.com/material-components/material-components-web-components/issues/480

mwc-slider has breaking changes in MDC web 4:

> slider: remove adapter methods appendTrackMarkers, removeTrackMarkers , setLastTrackMarkersStyleProperty , and add adapter method setTrackMarkers.

https://github.com/material-components/material-components-web-components/pulls?q=is%3Apr+is%3Aclosed deleted the 3 removed methods, and added a no-op for the new `setTrackMarkers` method, just to make tsc happy, but we presumably need to actually do something for `setTrackMarkers`.

e111077:
fixes #486.

Also cleaned up innerText implementation and cleaned up render function to look less ugly.

### DIFF
--- a/packages/slider/src/mwc-slider-base.ts
+++ b/packages/slider/src/mwc-slider-base.ts
@@ -19,7 +19,6 @@ import {MDCSliderAdapter} from '@material/slider/adapter.js';
 import MDCSliderFoundation from '@material/slider/foundation.js';
 import {html, property, query, TemplateResult} from 'lit-element';
 import {classMap} from 'lit-html/directives/class-map';
-import {repeat} from 'lit-html/directives/repeat.js';
 
 const {INPUT_EVENT, CHANGE_EVENT} = MDCSliderFoundation.strings;
 
@@ -79,11 +78,9 @@ export class SliderBase extends FormElement {
   })
   markers = false;
 
-  @property({type: Number}) private _numMarkers = 0;
-
   // TODO(sorvell) #css: needs a default width
   protected render() {
-    const {value, min, max, step, disabled, discrete, markers, _numMarkers} =
+    const {value, min, max, step, disabled, discrete, markers} =
         this;
     const hostClassInfo = {
       'mdc-slider--discrete': discrete,
@@ -93,14 +90,8 @@ export class SliderBase extends FormElement {
     let markersTemplate: TemplateResult|string = '';
 
     if (discrete && markers) {
-      const markerEls = repeat(
-          new Array(_numMarkers),
-          () => html`<div class="mdc-slider__track-marker"></div>`);
-
       markersTemplate = html`
-        <div class="mdc-slider__track-marker-container">
-          ${markerEls}
-        </div>`;
+        <div class="mdc-slider__track-marker-container"></div>`;
     }
     return html`
       <div class="mdc-slider ${classMap(hostClassInfo)}"
@@ -176,8 +167,17 @@ export class SliderBase extends FormElement {
           this.trackElement.style.setProperty(propertyName, value),
       setMarkerValue: (value: number) => this.pinMarker.innerText =
           value.toString(),
-      setTrackMarkers: (_step: number, _max: number, _min: number) => {
-        // TODO(aomarks) What are we supposed to do here?
+      setTrackMarkers: (step, max, min) => {
+        const stepStr = step.toLocaleString();
+        const maxStr = max.toLocaleString();
+        const minStr = min.toLocaleString();
+        // keep calculation in css for better rounding/subpixel behavior
+        const markerAmount = `((${maxStr} - ${minStr}) / ${stepStr})`;
+        const markerWidth = `2px`;
+        const markerBkgdImage = `linear-gradient(to right, currentColor ${markerWidth}, transparent 0)`;
+        const markerBkgdLayout = `0 center / calc((100% - ${markerWidth}) / ${markerAmount}) 100% repeat-x`;
+        const markerBkgdShorthand = `${markerBkgdImage} ${markerBkgdLayout}`;
+        this.trackMarkerContainer.style.setProperty('background', markerBkgdShorthand);
       },
       isRTL: () => getComputedStyle(this.mdcRoot).direction === 'rtl',
     };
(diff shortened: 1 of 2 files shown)


packages/slider/src/mwc-slider-base.ts
```diff
@@ -174,10 +170,21 @@ export class SliderBase extends FormElement {
...
+      setTrackMarkers: (step, max, min) => {
+        const stepStr = step.toLocaleString();
+        const maxStr = max.toLocaleString();
+        const minStr = min.toLocaleString();
+        // keep calculation in css for better rounding/subpixel behavior
+        const markerAmount = `((${maxStr} - ${minStr}) / ${stepStr})`;
+        const markerWidth = '2px';
+        const markerBkgdImage = `linear-gradient(to right, currentColor ${
+            markerWidth}, transparent 0)`;
+        const markerBkgdLayout = `0 center / calc((100% - ${markerWidth}) / ${
```
azakus: This is starting to get a little complicated. Can you put a comment in here about what this calculation is doing?

e111077: Sure, I hate it, but it's exactly what they do in the component. It basically draws the notches on the slider if you enable notches and steps. It's more performant, but ugly

e111077: done

e111077:
Re-requesting review as I refactored it a bit to not use `element.style.setProperty` to use `styleMap`. Additionally, added `{passive: true}` to some things that were giving us warnings

packages/slider/src/mwc-slider-base.ts
```diff
@@ -137,14 +146,18 @@ export class SliderBase extends FormElement {
       computeBoundingRect: () => this.mdcRoot.getBoundingClientRect(),
       getTabIndex: () => this.mdcRoot.tabIndex,
       registerInteractionHandler:
-          <K extends EventType>(type: K, handler: SpecificEventListener<K>) =>
-              this.mdcRoot.addEventListener(type, handler),
+          <K extends EventType>(type: K, handler: SpecificEventListener<K>) => {
+            const init = type === 'touchstart' ? {passive: true} : {};
```
azakus: This should be using `applyPassive()` from `@material/dom/events` like we use in ripple-directive: https://github.com/material-components/material-components-web-components/blob/master/packages/ripple/src/ripple-directive.ts#L17

packages/slider/src/mwc-slider-base.ts
```diff
@@ -137,14 +146,18 @@ export class SliderBase extends FormElement {
...
+            this.mdcRoot.addEventListener(type, handler, init);
+          },
       deregisterInteractionHandler:
           <K extends EventType>(type: K, handler: SpecificEventListener<K>) =>
               this.mdcRoot.removeEventListener(type, handler),
       registerThumbContainerInteractionHandler:
-          <K extends EventType>(type: K, handler: SpecificEventListener<K>) =>
-              this.thumbContainer.addEventListener(type, handler),
+          <K extends EventType>(type: K, handler: SpecificEventListener<K>) => {
+            const init = type === 'touchstart' ? {passive: true} : {};
```
azakus: `applyPassive()`

packages/slider/src/mwc-slider-base.ts
```diff
@@ -155,7 +168,7 @@ export class SliderBase extends FormElement {
           <K extends EventType>(type: K, handler: SpecificEventListener<K>) =>
               document.body.removeEventListener(type, handler),
       registerResizeHandler: (handler: SpecificEventListener<'resize'>) =>
-          window.addEventListener('resize', handler),
+          window.addEventListener('resize', handler, {passive: true}),
```
azakus: `applyPassive()`

packages/slider/src/mwc-slider-base.ts
```diff
@@ -127,6 +131,11 @@ export class SliderBase extends FormElement {
     </div>`;
   }
 
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.mdcFoundation.destroy();
```
azakus: Why destroy the foundation on `disconnectedCallback()`?

e111077: to remove all the body resize observers

packages/slider/src/mwc-slider-base.ts
```diff
@@ -147,15 +148,15 @@ export class SliderBase extends FormElement {
       getTabIndex: () => this.mdcRoot.tabIndex,
       registerInteractionHandler:
           <K extends EventType>(type: K, handler: SpecificEventListener<K>) => {
-            const init = type === 'touchstart' ? {passive: true} : {};
+            const init = type === 'touchstart' ? applyPassive() : {};
```
azakus: The fallback must be `undefined` to not force `capture` behavior on IE

packages/slider/src/mwc-slider-base.ts
```diff
@@ -147,15 +148,15 @@ export class SliderBase extends FormElement {
...
             this.mdcRoot.addEventListener(type, handler, init);
           },
       deregisterInteractionHandler:
           <K extends EventType>(type: K, handler: SpecificEventListener<K>) =>
               this.mdcRoot.removeEventListener(type, handler),
       registerThumbContainerInteractionHandler:
           <K extends EventType>(type: K, handler: SpecificEventListener<K>) => {
-            const init = type === 'touchstart' ? {passive: true} : {};
+            const init = type === 'touchstart' ? applyPassive() : {};
```
azakus: fallback to `undefined`

packages/slider/src/mwc-slider-base.ts
azakus: Seems like we either need an explicit removeEventListener wrapper in the foundation, or we'll need to recreate the foundation in `connectedCallback()`

e111077: re-initted in connected callback

PR #506, opened 2019-09-27
Update mwc-fab to MDC 4

### LINKED ISSUE #484: Update mwc-fab to MDC Web 4
We need to get MWC depending on MDC Web 4.0.0 canary.

The breaking changes are listed here: https://github.com/material-components/material-components-web/blob/v4.0.0-canary.e851d4f40.0/CHANGELOG.md#breaking-changes. Ripples for button, radio and fab now require an additional node.

The `canary` branch already exists with these new deps, and mwc-button has already been updated there (https://github.com/material-components/material-components-web-components/pull/481).

However, adding the new ripple node to mwc-fab causes the ripple to be much too large. Some additional changes seem to be needed.

e111077:
fixes #484.

Please note that it seems as if `mdc-fab` is broken on canary:

https://invited-emu.glitch.me/

### DIFF
--- a/packages/fab/src/mwc-fab-base.ts
+++ b/packages/fab/src/mwc-fab-base.ts
@@ -56,8 +56,9 @@ export class FabBase extends LitElement {
       <button
           class="mdc-fab ${classMap(classes)}"
           ?disabled="${this.disabled}"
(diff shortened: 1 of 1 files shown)


e111077:
b/141748664

PR #511, opened 2019-09-28
add missing dep to mwc-dialog

e111077:
it's already in the tsconfig, but forgot to add it to package.json. Does this need to go in changelog?


e111077:
@googlebot I fixed it.

aomarks:
> it's already in the tsconfig, but forgot to add it to package.json. Does this need to go in changelog?

Yeah, I'd probably mention it as a bug fix, since it looks like if you only installed this component, you would have gotten an error.

PR #534, opened 2019-10-03
Use display=block in demo to load Material Icons

aomarks:


aomarks:
Superseded by https://github.com/material-components/material-components-web-components/pull/546

PR #539, opened 2019-10-07
[mwc-textfield] call setValue when value changes

### LINKED ISSUE #515: mwc-textfield label not updated after value change
If the field is empty and the label is inside the field, setting `value` programmatically doesn't trigger the label repositioning.

Here's the reproduction : [https://mwc-textfield-update-value.glitch.me/](https://mwc-textfield-update-value.glitch.me/) (workaround included)

43081j:
Fixes #515

If we call `setValue` on the foundation when the value changes, all the base logic like positioning the floating label happens. there's already a check in the foundation for if the value is equal, so it won't double-update the input's value.

### DIFF
--- a/packages/textfield/src/mwc-textfield-base.ts
+++ b/packages/textfield/src/mwc-textfield-base.ts
@@ -250,6 +250,11 @@ export abstract class TextFieldBase extends FormElement {
     if (maxLengthBecameDefined || maxLengthBecameUndefined) {
       this.createFoundation();
     }
+
+    if (changedProperties.has('value') &&
+        changedProperties.get('value') !== undefined) {
+      this.mdcFoundation.setValue(this.value);
+    }
   }
 
   protected renderInput() {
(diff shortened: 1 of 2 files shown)


43081j:
@e111077 any idea why these tests fail? shouldn't be related to the change?

e111077:
It can be. Unfortunately the foundation does lots of stuff validation when value changes in the foundation model.

e111077:
Can you make sure that this `setValue` is not being called on the initial case when value is empty?

43081j:
looks like you were right @e111077!

`value` is a string (not an optional one) so it should be enough that we just check the previous value wasn't `undefined` as that should only ever happen during the constructor being executed (which means we're initialising).

e111077:
running internal tests

e111077:
passed and committed internally. IDK what's going on, but I'll wait until tomorrow to see if copybara is just backed up.

Issue #205, opened 2019-01-28
Button type submit


toastal:
I'm suprised this hasn't seemed to have come up, but you should be able to pass down button types like submit and reset. `<mdc-button type="submit" ... />`

dfreedm:
Related #205

salomvary:
Ugly workaround: `<mwc-button onclick="event.target.closest('form').submit()">`.

mohe2015:
I just found out that this workaround is bad if you have submit event listeners. You would need to use https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit which is not supported in enough browsers. So I think this should be prioritized because implementation doesn't seem that hard and it's really annoying. The only workaround is probably to use the onclick for submitting which probably doesn't work when pressing enter or so.

maicol07:
Any news on this?

fernandopasik:
Looking at the new implementation of the button component in 
https://github.com/material-components/material-web/blob/master/button/lib/button.ts

I noticed that `type` prop was missing again, is there a particular reason?

asyncLiz:
I think it makes sense to add `type` when we update components to be Form Associated Custom Elements (FACE)

fernandopasik:
Brilliant thank you! I noticed it's resolved
https://github.com/material-components/material-web/blob/main/button/lib/button.ts#L76

Issue #289, opened 2019-07-03
Hook components that should go into forms into Form Submission


dfreedm:
As referenced in https://github.com/material-components/material-components-web-components/issues/223#issuecomment-508284406, we should ensure that form-ish elements like `<mwc-radio>` and others participate in forms.

We should investigate if [Form Associated Custom Elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example) or the [FormData event](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-formdataevent-interface) are simpler to integrate and/or polyfill.

tschundler:
fwiw, an approach I've used is to wrap from elements in custom components and the webcomponent updates the value of the wrapped element. It could inject a hidden input into the wrapped light dom as a polyfill, or use one if it's already there to give the user more control.
(I'd like to switch some existing code to material, but can't since it isn't worth the dev effort for it to me more than a simple html/css change)

adrm:
What would be the downside of not using shadowDom for form-y components? Maybe you can use other css encapsulation technique that exposes the inner input to the light dom, because as they are right now, sadly it is very cumbersome to use this components in a form. This is the approach Ionic uses for their components, for example.

asyncLiz:
The downside is that element internals are unsafely exposed to external CSS styles. The purpose of shadow DOM is to provide protection and ensure that the elements are not affects by any external styles.

Luckily there are some great solutions for custom elements and forms as mentioned in this issue! The next step is implementing them. 

While you wait for support to be built-in, you can always hook up the components today using FormData events. Here's a great blog post with instructions on how to do that: https://web.dev/more-capable-form-controls/#event-based-api

adrm:
Thanks for the explanation. In the meantime, I am copying the value on changes to a hidden input so the form can access it. I hope we can all use FormData events soon but it seems that right now they are not supported by Safari and no easy polyfill is possible (https://caniuse.com/#search=formdata%20event). Great work you are doing over here. Thanks a lot!

fernandopasik:
If the choice is to implement [Form Associated Custom Elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example) there's a polyfill that might be worth investigating
https://github.com/calebdwilliams/element-internals-polyfill

YonatanKra:
@dfreedm 
[...]

PR #540, opened 2019-10-08
Add button type property in mwc-button

jpnathan:
Issue: Button type submit #205

### DIFF
--- a/packages/button/src/mwc-button-base.ts
+++ b/packages/button/src/mwc-button-base.ts
@@ -14,7 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import {HTMLElementWithRipple} from '@material/mwc-base/form-element';
+import {HTMLElementWithRipple } from '@material/mwc-base/form-element';
 import {ripple} from '@material/mwc-ripple/ripple-directive.js';
 import {html, LitElement, property, query} from 'lit-element';
 import {classMap} from 'lit-html/directives/class-map';
@@ -31,6 +31,8 @@ export class ButtonBase extends LitElement {
   @property({type: Boolean, reflect: true}) disabled = false;
 
   @property({type: Boolean}) trailingIcon = false;
+
+  @property({type: String}) type = 'submit';
 
   @property() icon = '';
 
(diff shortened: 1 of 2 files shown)


azakus:
Related to #289

packages/button/src/mwc-button-base.ts
```diff
@@ -32,6 +32,8 @@ export class ButtonBase extends LitElement {
 
   @property({type: Boolean}) trailingIcon = false;
 
+  @property({type: String}) type = 'submit';
```
fernandopasik, line 35: I think the in the html spec the default type is `button`

Also might be good to add a typescript string literal type for this
`type ButtonType = 'button' | 'submit' | 'reset'`

resulting in 
`@property({type: String}) type: ButtonType = 'button';`

packages/button/src/mwc-button-base.ts
```diff
@@ -14,7 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import {HTMLElementWithRipple} from '@material/mwc-base/form-element';
+import {HTMLElementWithRipple } from '@material/mwc-base/form-element';
```
fernandopasik, line 17: Please remove this unnecessary white space

fernandopasik:
> Related to #289

@dfreedm I know that is related to the form submit ticket, but why should that block adding this feature?

packages/button/src/mwc-button-base.ts
salomvary: FYI the default type in HTML is `submit`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type

fernandopasik: sorry about that, yes `submit` is the default

asyncLiz:
What does adding this feature without custom element form submission accomplish? 

An `<mwc-button type="submit">` inside a `<form>` element will not submit it. I'd be cautious about adding `type="submit"` which could imply to users that form submission works when it doesn't.

fernandopasik:
Makes sense @asyncLiz, thanks. We'll need to wait on the result of #289

asyncLiz:
Closing as obsolete

fernandopasik:
The type attribute wasn't added to button? Is it resolved in some other way?

asyncLiz:
Sorry about that! Was going through and cleaning up a lot of old issues :)

fernandopasik:
would it make sense to rebase this PR and try to get it ready to merge?

asyncLiz:
> would it make sense to rebase this PR and try to get it ready to merge?

If you can rebase it for MWC (skip M3), I'll review it :)

asyncLiz:
Obsolete with M3. See https://github.com/material-components/material-web/issues/3941 for upcoming support

PR #546, opened 2019-10-08
Remove mwc-icon-font.js

aomarks:
Let's just have the one way to load fonts, as documented in our README.
Updated all the demos to use the same recommendation.


aomarks:
I updated all the files in https://mwc-demos.glitch.me manually

PR #547, opened 2019-10-09
Move character-counter-directive out of its own directory

aomarks:
In Google, every directory containing TypeScript files must be a directory with a BUILD file (a package), and packages cannot contain dashes. We could handle this case with a transformation, but it seems
simpler to just keep a flatter layout upstream.


e111077:
Gosh I love our modular code and how this only had to be fixed here

PR #549, opened 2019-10-09
Textfield fill and ink styling

### LINKED ISSUE #460: there is no way to color the helper and input text
in element `mwc-textfield` and `mwc-textarea`, I can not find a way to set `helper text color`, `input text color`, `background color` and `unfocusd outlined outline color`

### LINKED ISSUE #544: expose method to set background-color on mwc-textfield
attempting this fix in mdc...

e111077:
Fixes #460.
Fixes #544.

Refactored textfield styling to make more conceptual sense and collapse common selectors. Additionally, added the following css variables:

## mwc-textfield

| Name                                              | Default               | Description
| ------------------------------------------------- | --------------------- |------------
| `--mdc-text-field-fill-color`                     | filled: `rgb(245, 245, 245)`, outlined: `transparent` | Color of the textfield's background fill.
| `--mdc-text-field-disabled-fill-color`            | filled: `rgb(250, 250, 250)`, outlined: `transparent` | Color of the textfield's background fill when disabled.
| `--mdc-text-field-ink-color`                      | `rgba(0, 0, 0, 0.87)` | Color of the input text.
| `--mdc-text-field-label-ink-color`                | `rgba(0, 0, 0, 0.6)`  | Color of the non-focused floating label, helper text, char counter, and placeholder.
| `--mdc-text-field-disabled-ink-color`             | `rgba(0, 0, 0, 0.37)` | Color of the input text, the floating label, helper text, char counter, and placeholder of a disabled textfield.

## mwc-notched-outline
| Name                                           | Default       | Description
| ---------------------------------------------- | ------------- |------------
| `--mdc-notched-outline-border-color`           | none          | Sets the border / outline color. **(NOTE only added missing documentation for this one)**
| `--mdc-notched-outline-background-color`       | none          | Sets the background color inside the styled borders.

### DIFF
--- a/packages/textfield/README.md
+++ b/packages/textfield/README.md
@@ -213,12 +213,8 @@ Inherits CSS Custom properties from:
 | `--mdc-text-field-outlined-idle-border-color`     | `rgba(0, 0, 0, 0.38)` | Color of the outlined textfield's  outline when idle.
 | `--mdc-text-field-outlined-hover-border-color`    | `rgba(0, 0, 0, 0.87)` | Color of the outlined textfield's outline when hovering.
 | `--mdc-text-field-outlined-disabled-border-color` | `rgba(0, 0, 0, 0.06)` | Color of the outlined textfield's outline when disabled.
-| `--mdc-text-field-fill-color`                     | filled: `rgb(245, 245, 245)`, outlined: `transparent` | Color of the textfield's background fill.
-| `--mdc-text-field-outlined-fill-color`            | `transparent`         | Overrides `--mdc-text-field-fill-color` only for outlined textfields.
-| `--mdc-text-field-filled-fill-color`              | `rgb(245,245,245)`    | Overrides `--mdc-text-field-fill-color` only for filled (non-outlined) textfields.
-| `--mdc-text-field-disabled-fill-color`            | filled: `rgb(250, 250, 250)`, outlined: `transparent` | Color of the textfield's background fill when disabled.
-| `--mdc-text-field-outlined-disabled-fill-color`   | `rgba(250, 250, 250)` | Overrides `--mdc-text-field-disabled-fill-color` only for outlined textfields.
-| `--mdc-text-field-filled-disabled-fill-color`     | `transparent`         | Overrides `--mdc-text-field-disabled-fill-color` only for filled (non-outlined) textfields.
+| `--mdc-text-field-fill-color`                     | `rgb(245, 245, 245)`  | Color of the textfield's background fill (non-outlined).
+| `--mdc-text-field-disabled-fill-color`            | `rgb(250, 250, 250)`  | Color of the textfield's background fill (non-outlined) when disabled.
 | `--mdc-text-field-ink-color`                      | `rgba(0, 0, 0, 0.87)` | Color of the input text.
 | `--mdc-text-field-label-ink-color`                | `rgba(0, 0, 0, 0.6)`  | Color of the non-focused floating label, helper text, char counter, and placeholder.
 | `--mdc-text-field-disabled-ink-color`             | `rgba(0, 0, 0, 0.37)` | Color of the input text, the floating label, helper text, char counter, and placeholder of a disabled textfield.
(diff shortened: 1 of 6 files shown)


e111077:
my bad for not making the reformat a different commit >.<

packages/textfield/README.md
```diff
@@ -213,6 +213,15 @@ Inherits CSS Custom properties from:
 | `--mdc-text-field-outlined-idle-border-color`     | `rgba(0, 0, 0, 0.38)` | Color of the outlined textfield's  outline when idle.
 | `--mdc-text-field-outlined-hover-border-color`    | `rgba(0, 0, 0, 0.87)` | Color of the outlined textfield's outline when hovering.
 | `--mdc-text-field-outlined-disabled-border-color` | `rgba(0, 0, 0, 0.06)` | Color of the outlined textfield's outline when disabled.
+| `--mdc-text-field-fill-color`                     | filled: `rgb(245, 245, 245)`, outlined: `transparent` | Color of the textfield's background fill.
```
aomarks: Is it within the spec to have an outlined textfield without a transparent background?

packages/textfield/README.md
```diff
@@ -213,6 +213,15 @@ Inherits CSS Custom properties from:
...
+| `--mdc-text-field-filled-fill-color`              | `rgb(245,245,245)`    | Overrides `--mdc-text-field-fill-color` only for filled (non-outlined) textfields.
+| `--mdc-text-field-disabled-fill-color`            | filled: `rgb(250, 250, 250)`, outlined: `transparent` | Color of the textfield's background fill when disabled.
+| `--mdc-text-field-outlined-disabled-fill-color`   | `rgba(250, 250, 250)` | Overrides `--mdc-text-field-disabled-fill-color` only for outlined textfields.
+| `--mdc-text-field-filled-disabled-fill-color`     | `transparent`         | Overrides `--mdc-text-field-disabled-fill-color` only for filled (non-outlined) textfields.
+| `--mdc-text-field-ink-color`                      | `rgba(0, 0, 0, 0.87)` | Color of the input text.
+| `--mdc-text-field-label-ink-color`                | `rgba(0, 0, 0, 0.6)`  | Color of the non-focused floating label, helper text, char counter, and placeholder.
+| `--mdc-text-field-disabled-ink-color`             | `rgba(0, 0, 0, 0.37)` | Color of the input text, the floating label, helper text, char counter, and placeholder of a disabled textfield.
```
aomarks: Does this mean that when the input is disabled, you can't have different colors for the input text and the label text (it sounds like when it's enabled there's both `--mdc-text-field-ink-color` and `--mdc-text-field-label-ink-color`, but when it's disabled there's only `--mdc-text-field-disabled-ink-color`).

packages/textfield/README.md
e111077: nothing in the spec, but reading over the mdc docs, this is not a thing they expose. Removing it.

packages/textfield/README.md
e111077: So, when not disabled, the label and the text are the same color, but they are at different opacities. when they are disabled, they are all the same color.

I can instead ask for an rgb list of:

`--mdc-text-field-ink-color: 128, 128, 128;` and then apply

```
input {
  color: rgba(var(--mdc-text-field-ink-color), .87)
}

label,
input::placeholder,
.char-counter,
.helper-text {
  color: rgba(var(--mdc-text-field-ink-color), .6);
}
```

I'd just have to change the documentation to accept a comma rgb value instead of a color or a hex color.

e111077: this is how it works with the sass:

https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/_variables.scss

aomarks: I dont really have an opinion about how this should work, your call.

e111077: We are considering a bunch of more-extensible fixes for the near future. I think this is fine for now since a more-structural change needs to happen for the --surface and --on-surface nirvana

PR #554, opened 2019-10-11
Prepare to release 0.10.0

aomarks:


e111077:
looks like there's a breaking change in the deps

aomarks:
> looks like there's a breaking change in the deps

Filed https://github.com/material-components/material-components-web-components/issues/555 and reverted for this PR.

PR #556, opened 2019-10-14
Dialog title -> heading + aria fix

e111077:
`Dialog.title` renamed to `Dialog.heading` and `--mdc-dialog-title-ink-color` renamed to `--mdc-dialog-heading-ink-color` as it caused clashes with `HTMLElement.prototype.title` causing a tooltip to appear everywhere.

Also quick fix to aria-labeledby and aria-describedby. Checked and works with mac voiceover


e111077:
why didn't this merge?

e111077:
oh it was submitted internally just now. oops. I'll give it a wait

PR #564, opened 2019-10-16
Fix checkbox ripple visibility when focused while being unchecked.

sorvell:
The ripple directive needs to be told to watch the native input element for focus/blur events.


aomarks:
From Travis: `ERROR:Project is not formatted. Please run "npm run format".`

aomarks:
Travis is going to be red right now due to an MDC web change that we need to pull in here via a new dependency.

aomarks:
> Travis is going to be red right now due to an MDC web change that we need to pull in here via a new dependency.

We're a little stuck here in GitHub until that breaking MDC web change is out, but everything should be fine in Google, so let's pull this into Google early (usually we would wait until Travis is green before pulling into Google).

PR #592, opened 2019-10-25
Update README.md

davie-robertson:
Replaced incorrect reference to snackbar with button


aomarks:
Thanks for catching this!

PR #607, opened 2019-11-05
fix ripple shadow root

e111077:
ripple won't be able to register clicks if the direct parent is a shadow root. This fixes this.


packages/ripple/src/test/mwc-ripple.test.ts
```diff
@@ -67,5 +67,23 @@ suite('mwc-ripple', () => {
...
+
+      root.appendChild(ripple);
+
+      await ripple.updateComplete;
+
+      assert.instanceOf(internals.interactionNode, HTMLElement);
+      assert(internals.interactionNode === container);
+
+      document.body.removeChild(container);
+    })
```
azakus: Missing semicolon breaking lint

PR #609, opened 2019-11-06
Slider scroll fix

### LINKED ISSUE #574: Slider does not work inside a Dialog

**I'm submitting a:**

- [x] bug report
- [ ] feature request

**What OS are you using?**

Kubuntu 19.10

Chrome Version 77.0.3865.120 (Official Build) (64-bit)

**Current behavior:**

The Slider is not moveable by mouse, but the value is changeable by keyboard, when used inside a dialog. It works after the browser window is resized.

**Expected behavior:**

**Steps to reproduce:**

Just put a slider into a dialog

**Related code:**

```js
import { html, css, LitElement } from 'lit-element';
import '@material/mwc-dialog';
import '@material/mwc-slider';

export class PageOne extends LitElement {
  static get styles() {
    return css`
      :host { display: block; padding: 25px; }
    `;
  }

  firstUpdated(changedProperties) {
    this._dialog = this.shadowRoot.getElementById('dialog');
  }

  __show() {
    this._dialog.open = true;
  }

  render() {
    return html`
      <button @click=${this.__show}>show dialog</button>
      <mwc-dialog id="dialog" title="Dialog">
        <form>
          <mwc-slider discrete="" markers="" min="0" max="5"></mwc-slider>
        </form>
      </mwc-dialog>
    `;
  }
}
```

**Other information:**

e111077:
Fixed issue where slider when scrolled horizontally or when resized would not slide properly by calling layout (which gets bounding client rect) on touchstart or mousedown.

Also changelog for this and my other icon-button, ripple and snackbar PRs

Fixes #574

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - CSS styling options to `mwc-textfield`
 - README for `mwc-drawer`
 - Demo for `mwc-drawer` without a header in the drawer
+- `--mdc-icon-button-size` and `--mdc-icon-size` to `mwc-icon-button`
 
 ### Changed
 
@@ -30,6 +31,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed issue where slider when resized or scrolled will not respond to touch
   as expected.
 - Fixed issue where `mwc-ripple` would not ripple when parent was a shadow root
+- **BREAKING:VISUAL** Fixed sizing of the `mwc-icon-button` in `mwc-snackbar`
 
 ## [0.10.0] - 2019-10-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `mwc-dialog` will now search its flattened distributed nodes and their trees
   for a focusable element.
 - **BREAKING** `mwc-slider` now emits bubbling and composed `input` and `change`
-  events from `MDCSlider:input` and `MDCSlider:change`.
+  events instead of `MDCSlider:input` and `MDCSlider:change`.
 
 ### Fixed
 - Fixed checkbox ripple visibility when focused while being unchecked.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Updated material dependencies to `4.0.0-canary.735147131.0`.
 - `mwc-dialog` will now search its flattened distributed nodes and their trees
   for a focusable element.
+- **BREAKING** `mwc-slider` now emits bubbling and composed `input` and `change`
+  events from `MDCSlider:input` and `MDCSlider:change`.
 
 ### Fixed
 - Fixed checkbox ripple visibility when focused while being unchecked.
 - Fixed app content not being expanded inside drawer.
+- Fixed issue where slider when resized or scrolled will not respond to touch
+  as expected.
 
 ## [0.10.0] - 2019-10-11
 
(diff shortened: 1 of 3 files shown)


packages/slider/src/mwc-slider-base.ts
```diff
@@ -18,11 +18,12 @@ import {applyPassive} from '@material/dom/events.js';
 import {addHasRemoveClass, EventType, FormElement, observer, SpecificEventListener} from '@material/mwc-base/form-element.js';
 import {MDCSliderAdapter} from '@material/slider/adapter.js';
 import MDCSliderFoundation from '@material/slider/foundation.js';
-import {html, property, query, TemplateResult} from 'lit-element';
+import {eventOptions, html, property, query, TemplateResult} from 'lit-element';
 import {classMap} from 'lit-html/directives/class-map';
 import {styleMap} from 'lit-html/directives/style-map';
 
-const {INPUT_EVENT, CHANGE_EVENT} = MDCSliderFoundation.strings;
```
aomarks: What happened here? Why can't we use the MDC constants?

packages/slider/src/mwc-slider-base.ts
```diff
@@ -226,6 +231,7 @@ export class SliderBase extends FormElement {
     };
   }
 
+  @eventOptions({capture: true, passive: true})
```
aomarks: Can you add a comment about why we need to call layout on mousedown and touchmove? Seems like that could be expensive because, isn't touchmove going to get fired a lot when a user is interacting with the slider? I'm not sure I understand the original problem this is fixing very well.

packages/slider/src/mwc-slider-base.ts
e111077: we don't use the MDC constants because they are prefixed `MDCSlider:change` or `MDCSlider:input` which are leaky and unnecessary since the element affords mdc slider

packages/slider/src/mwc-slider-base.ts
e111077: meant to do touchstart

e111077: fixed

e111077:
also added change fixing scroll x. I seemed to have previously only fixed the initial display being none issue (linked in issue)

packages/slider/src/mwc-slider-base.ts
aomarks: I think it still needs a comment.

packages/slider/src/mwc-slider-base.ts
aomarks: For drawer we use MDCDrawer prefixed events: https://github.com/material-components/material-components-web-components/tree/master/packages/drawer#events -- why is this different?

> which are leaky and unnecessary since the element affords mdc slider

Sorry can you rephrase? Not quite sure I understand what this means.

e111077: Oops I thought we did not use prefixed events for Drawer.

I think we should use `input` and `change` for slider though to emulate `<input type="range">`.

And by leaky, I mean that we are leaking the implementation details of this slider by firing an event that would not emulate the expected events from native `input`. The user would have to know about our internal implementation just to listen to a change and input event.

This is why we got rid of the prefixes in [dialog](https://github.com/material-components/material-components-web-components/tree/master/packages/dialog#events)

CHANGELOG.md
```diff
@@ -21,10 +22,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Updated material dependencies to `4.0.0-canary.735147131.0`.
 - `mwc-dialog` will now search its flattened distributed nodes and their trees
   for a focusable element.
+- **BREAKING** `mwc-slider` now emits bubbling and composed `input` and `change`
+  events from `MDCSlider:input` and `MDCSlider:change`.
```
aomarks: did you mean "events instead of" instead of "events from"?

packages/slider/src/mwc-slider-base.ts
aomarks: I still don't see a comment.

e111077: weird

e111077: thanks for the catch!

CHANGELOG.md
e111077: i do now!

PR #610, opened 2019-11-06
search flattened distributed for focus

e111077:
Dialog will now search its distributed nodes' trees for focus. Needed for the following pattern:

```html
<my-dialog>
  <input dialogInitialFocus>
  #shadowRoot
    <mwc-dialog>
      <slot></slot>
    </mwc-dialog>
</my-dialog>
```

### DIFF
--- a/packages/dialog/src/mwc-dialog-base.ts
+++ b/packages/dialog/src/mwc-dialog-base.ts
@@ -130,28 +130,30 @@ export class DialogBase extends BaseElement {
     }
 
     // if not in light dom, search each flattened distributed node.
-    const content = this.contentSlot as HTMLSlotElement;
     const primary = this.primarySlot as HTMLSlotElement;
     const secondary = this.secondarySlot as HTMLSlotElement;
+    const content = this.contentSlot as HTMLSlotElement;
 
-    const contentNodes = content.assignedNodes({flatten: true});
+
     const primaryNodes = primary.assignedNodes({flatten: true});
-    const secondaryNodes = secondary.assignedNodes({flatten: true});
-
-    const initFocusElement = this.searchNodeTreesForAttribute(
-        contentNodes, this.initialFocusAttribute);
-    if (initFocusElement) {
-      return initFocusElement;
-    }
-
     const primaryFocusElement = this.searchNodeTreesForAttribute(
         primaryNodes, this.initialFocusAttribute);
     if (primaryFocusElement) {
       return primaryFocusElement;
     }
+
+    const secondaryNodes = secondary.assignedNodes({flatten: true});
     const secondaryFocusElement = this.searchNodeTreesForAttribute(
         secondaryNodes, this.initialFocusAttribute);
-    return secondaryFocusElement;
+    if (secondaryFocusElement) {
+      return secondaryFocusElement;
+    }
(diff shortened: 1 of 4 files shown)


packages/dialog/src/mwc-dialog-base.ts
```diff
@@ -118,7 +120,58 @@ export class DialogBase extends BaseElement {
...
+    // if not in light dom, search each flattened distributed node.
+    const content = this.contentSlot as HTMLSlotElement;
+    const primary = this.primarySlot as HTMLSlotElement;
+    const secondary = this.secondarySlot as HTMLSlotElement;
+
+    const contentNodes = content.assignedNodes({flatten: true});
+    const primaryNodes = primary.assignedNodes({flatten: true});
+    const secondaryNodes = secondary.assignedNodes({flatten: true});
+
+    const initFocusElement = this.searchNodeTreesForAttribute(
```
aomarks: Can you re-order these lines so that you don't run the query selector and assignNodes calls for primary and secondary when you don't need to?

packages/dialog/src/mwc-dialog-base.ts
```diff
@@ -118,7 +120,58 @@ export class DialogBase extends BaseElement {
...
+    for (const node of nodes) {
+      if (!(node instanceof HTMLElement)) {
+        continue;
+      }
+
+      if (node.hasAttribute(attribute)) {
+        return node;
+      } else {
+        const selection = node.querySelector(`[${attribute}]`);
+        if (selection instanceof HTMLElement) {
```
aomarks: Is it possible for these nodes not to be HTMLElements?

e111077: `querySelector` returns `null|Node`. So this both checks whether it's null and casts it to HTMLElement at the same time

aomarks: Yeah, but the instanceof check happens at runtime so there is a cost vs a cast. Is there something that could have the attribute but not be an HTMLElement (genuine question -- an SVG element or something? would that even work?)

e111077: good point

packages/dialog/src/mwc-dialog-base.ts
e111077: rearranged to primary -> secondary -> content since primary and secondary should have shorter trees. don't want to do work unless necessary

packages/dialog/src/mwc-dialog-base.ts
```diff
@@ -118,7 +120,60 @@ export class DialogBase extends BaseElement {
...
+    // only search light DOM. This typically handles all the cases
+    const lightDomQs = this.querySelector(initFocusSelector);
+
+    if (lightDomQs) {
+      return lightDomQs as HTMLElement;
+    }
+
+    // if not in light dom, search each flattened distributed node.
+    const primary = this.primarySlot as HTMLSlotElement;
+    const secondary = this.secondarySlot as HTMLSlotElement;
```
aomarks: This property access calls the query selector, so I think you should move this, and the content one, down to where they are used too, so that we can short-circuit them.

Issue #380, opened 2019-08-17
Radio, Switch, Formfield alignment issues


aomarks:
I tried what seemed to be a pretty straightforward layout of some selection controls using flex:

```html
<style>
  section {
    display: flex;
    flex-direction: column;
  }
</style>

<section>
  <mwc-formfield label="Checkbox"><mwc-checkbox></mwc-checkbox></mwc-formfield>
  <mwc-formfield label="Checkbox"><mwc-checkbox></mwc-checkbox></mwc-formfield>

  <mwc-formfield label="Radio"><mwc-radio></mwc-radio></mwc-formfield>
  <mwc-formfield label="Radio"><mwc-radio></mwc-radio></mwc-formfield>

  <mwc-formfield label="Switch"><mwc-switch></mwc-switch></mwc-formfield>
  <mwc-formfield label="Switch"><mwc-switch></mwc-switch></mwc-formfield>
</section>
```

Which renders like this:

<img width="147" alt="Screen Shot 2019-08-16 at 4 59 58 PM" src="https://user-images.githubusercontent.com/48894/63204115-7b94b680-c049-11e9-86ca-96b4b89a186e.png">

Here it is at 2X scale, with lines added to show some of the issues:

<img width="294" alt="Screen Shot 2019-08-16 at 4 59 58 PM 2" src="https://user-images.githubusercontent.com/48894/63204116-7cc5e380-c049-11e9-925d-7b294b2c9d30.png">

1) The switch label is not horizontally aligned with the radio and checkbox labels (green line).
2) The radio label is not vertically aligned with the radio button (red line).
3) The switch label is not vertically aligned with the switch (red line).
4) Switches have significantly less vertical margin than checkboxes or radios (pink line).
5) Checkboxes and radios have similar vertical margin, but not quite the same (pink line).

I think we should test this out using core MDC, to see if the same issues exist.

It's possible this is how MDC is designed, in which case we need to provide some additional help for users to get a decent layout: a helper element, some exported styles, or at the very least some documentation.

(But my initial reaction, whatever the cause, was that it seems like these components should tend to align themselves naturally by default, by having consistent sizes and margins).

dfreedm:
Related #1037 for Switch

asyncLiz:
Notably the horizontal alignment was solved in switch by setting the `:host` to `display: inline-flex;`.

asyncLiz:
[...]

PR #612, opened 2019-11-06
mwc-radio vertical alignment

### LINKED ISSUE #611: [mwc-radio & mwc-formfield] middle alignement issue
When using `mwc-radio` in a `mwc-formfield` the label is not properly aligned with the radio button.

reproduction : [https://mwc-radio-and-mwc-formfield-alignment.glitch.me/](https://mwc-radio-and-mwc-formfield-alignment.glitch.me/) (with hotfix)

vdegenne:
Fixes #611

### DIFF
--- a/packages/radio/src/mwc-radio.scss
+++ b/packages/radio/src/mwc-radio.scss
@@ -20,6 +20,5 @@ limitations under the License.
 @include mdc-ripple-common(mdc-feature-all());
 
 :host {
-  display: inline-flex;
   outline: none;
 }
--- a/packages/radio/src/mwc-radio.scss
+++ b/packages/radio/src/mwc-radio.scss
@@ -20,6 +20,6 @@ limitations under the License.
 @include mdc-ripple-common(mdc-feature-all());
 
 :host {
-  display: inline-block;
+  display: inline-flex;
   outline: none;
 }
--- a/packages/radio/src/mwc-radio.scss
+++ b/packages/radio/src/mwc-radio.scss
@@ -20,5 +20,10 @@ limitations under the License.
 @include mdc-ripple-common(mdc-feature-all());
 
 :host {
+  display: inline-block;
   outline: none;
+}
+
+.mdc-radio {
+  vertical-align: bottom;
 }


e111077:
internal tests are being manually run

vdegenne:
BTW I am not sure if flexing or blocking the radio element is the way to go as they should naturally be inline right ? So I found out a better way to prevent undesired alignment + keeping linearity with the checkbox element is to completely remove the display directive in the :host block of mwc-radio. mwc-checkbox has no display directive.

e111077:
actually, i am leaning towards leaving these as `display: inline-block` as that is what `input[type=radio]` does by default. This should be fixed in formfield.

e111077:
do you want to tackle this in this PR or in a different one?

vdegenne:
Is it really compulsory to follow the design of the original radio button ? For

1 ) Users that intend to use this type of components are not necessarily expecting the exact same blueprint.

2 ) Removing the `display` directive has one consequence : when the element is used in another `block` container its `display` property is automatically changed to `inline` and when the element is used in a `flex` element (e.g. `mwc-formfield`) its `display` property is automatically changed to `block`.

Even if users expect `mwc-radio` to match the exact same design as of the original radio element they won't experience any side effects for the element is inlined by default and squared blocked in a `mwc-formfield`. I am almost positive that the `display` property of the original radio element was set to `inline-block` so the element is inlined and it prevents the inner design to render differently than a square as it is the case for the current state of this element.

If it's a strict rule to keep the display to `inline-block`, we should understand why the element is rendering an extra space at the bottom when used in a formfield.

e111077:
Hello, sorry for the wait, I may have misunderstood your previous comment, but here is my response and please help clarify where we may be having a misunderstanding.

1. Since we are making low-level custom form elements, we want to attempt parity with their native analogs as best we can. It provides affordances to the user since users will be used to dealing with the native version of radio.

2. I may have misunderstood here, but placing an element with `display: inline` inside of a flex still sets the display to `block`. Here is an [example](https://stackblitz.com/edit/vanilla-webcomponent-p2lgrx?file=index.html).

I believe this should be fixed in `mwc-formfield` where it should have a selector of `::slotted(*) { display: flex; }` or `inline-flex`.

e111077:
Alternatively, if we want to fix this here, we can do what we did in `checkbox` and set `.mwc-radio { vertical-align: bottom; }`

e111077:
cc @aomarks and #380

vdegenne:
Hello,
I think there are no misunderstandings. What I meant (and relying on your stackblitz) is that removing the `display` directive is not changing how the element behaves externally : the element is inline and behaves as a block in a flex context. That's the original behavior of HTML implementation.
Which makes me think the motivation behind `display: inline-block` of the original radio button is for internal reason only (maybe to force the element to be a square as intended). Therefore I think removing the `display` directive is not going to have any drawbacks on its intended usage.

Of course this is speculative and I am just explaining why I thought removing `display` was a better proposal. There is another reason why I think modifying `mwc-formfield` is not really going to solve this particular problem.

Please see my updated demo : https://mwc-radio-and-mwc-formfield-alignment.glitch.me/
The grey box is a simple div element, the `mwc-radio` background is red and the div background at the root of `mwc-radio` shadow-dom is blue. There is a little bit of red left at the bottom as you could see.
Updating `mwc-formfield` is not going to fix this glitch and one would expect the element to be a square in any context right ?

From that I see only few ways to solve this properly :
- Find what is causing this extra gap
- remove `display` or set the display to `inline-flex`
- implementing a mechanism inside the element to force it to be square

(note: Changing `vertical-align` of `mwc-radio` in the demo has no visible effects)

Please tell me if we understand the same things now and sorry but I have not enough understanding of the source code to provide an appropriate PR, I am just trying to share my insight on the issue.

e111077:
Thank you so much for the clarification. The glitch demo makes me completely understand what you're getting across now. Thank you again for the deep investigation.

And sorry, I had a typo in my vertical-align suggestion. I meant setting `.mdc-radio` not `.mwc-radio`. This is referring to the [root element](https://github.com/material-components/material-components-web-components/blob/master/packages/radio/src/mwc-radio-base.ts#L154) (`<div class="mdc-radio ...">`) not the host (`:host`). This is what is done by the MDC library's SASS in checkbox, but, for some reason, this is not done in radio.

In your example, you can see if we set `.mdc-radio { vertical-align: bottom; }` it removes the red while keeping the same API surface of `inline-block` as `<input type="radio">`.

Again, please let me know if you would like to take this fix on. It would mean adding the css rule mentioned above in the same file that you had edited in this PR. If not, I don't mind doing it myself, but I just want to give you the chance to get the contribution credit in the commit history.

vdegenne:
Fixes #611 

I can't thank enough the Polymer and the Material team that gave me so much for my projects to be made possible, this is the least I can contribute.

Also should I open an issue in `material-components-web` repository to report about this tweak we had to make and reference this discussion ?

e111077:
Cool, I just need to kick our CI again. also our build is broken so it will fail again, but there was a weird issue in it

vdegenne:
Does the issue originate from this particular PR ?

e111077:
No it does not. Our release process is a bit slow. Sorry

PR #613, opened 2019-11-07
Slider readme & rename discrete prop

### LINKED ISSUE #582: Document Slider

e111077:
Fixes #582 

Make readme for slider. Also, I renamed `Slider.discrete` to `Slider.pin` since it seems as if that's really the only thing it did and `discrete` could be inferred from `Slider.step !== 0`. This is a breaking change and noted in the changelog


packages/slider/README.md
```diff
@@ -1,64 +1,124 @@
...
+> major changes until 1.0 release.
 
-A [Material Components](https://material.io/components/) icon implementation using [Web Components](https://www.webcomponents.org/introduction)
+Sliders allow users to make selections from a range of values.
 
-## Getting started
+**Note: vertical sliders and range (multi-thumb) sliders are not supported, due to their absence from the material design spec.**
 
- * The easiest way to try out mwc-slider is to use one of these online tools:
+<img src="images/standard.png" width="397px">
```
aomarks: This image doesn't exist.

Also I think it's really *height* that we should specify in the img tags, not width, to prevent jank.

packages/slider/README.md
```diff
@@ -1,64 +1,124 @@
...
 
-> :warning: These components are a work in progress. They are pre-release and should be considered experimental, as they may undergo major changes before release. We are experimenting with alternate architectures and approaches with the goal of allowing us to bring the most correct and optimal implementation of Material components to the widest possible audiences. Visible progress may be slow, as this research is across teams and repositories so is not consistently reflected in commits to this codebase. :warning:
+> IMPORTANT: The Material Web Components are a work in progress and subject to
+> major changes until 1.0 release.
 
-A [Material Components](https://material.io/components/) icon implementation using [Web Components](https://www.webcomponents.org/introduction)
+Sliders allow users to make selections from a range of values.
 
-## Getting started
+**Note: vertical sliders and range (multi-thumb) sliders are not supported, due to their absence from the material design spec.**
```
aomarks: Not sure we need this disclaimer.

packages/slider/README.md
```diff
@@ -1,64 +1,124 @@
...
-  1. Install the Polymer CLI:
+### Properties/Attributes
 
-      ```npm i -g polymer-cli```
+| Name      | Type      | Default | Description
+| --------- | --------- |-------- | -----------
+| `value`   | `number`  | `0`     | Current value of the slider.
+| `min`     | `number`  | `0`     | Minimum value of the slider.
+| `max`     | `number`  | `100`   | Maximum value of the slider.
+| `step`    | `number`  | `0`     | Sets the step of the slider making it discrete. Set to 0 to make continuous. Note: if `step` is greater than 0 but less than 1, it will be rounded up to 1.
```
aomarks: Won't any number be rounded up, not just 0.X?

Can you clarify more explicitly in these descriptions any limitations with integers vs floats? Are fractional numbers supported in continuous mode, or does slider only work with integers?

packages/slider/README.md
```diff
@@ -1,64 +1,124 @@
...
+| `pin`     | `bookean` | `false` | Shows the thumb pin on a discrete slider.
+| `markers` | `bookean` | `false` | Shows the marker on the slider's track of a discrete slider.
```
aomarks: `boolean`

aomarks: Maybe "Shows tick marks for each step on the track when the slider is discrete."

packages/slider/README.md
```diff
@@ -1,64 +1,124 @@
...
-      ```polymer serve```
+| Name     | Description
+| -------- | -------------
+| `layout() => void` | Updates the internal size model. May be required if the slider is resized or repositioned horizontally.
 
-  > mwc-slider is published on [npm](https://www.npmjs.com/package/@material/mwc-slider) using JavaScript Modules.
-  This means it can take advantage of the standard native JavaScript module loader available in all current major browsers.
-  >
-  > However, since mwc-slider uses npm convention to reference dependencies by name, a light transform to rewrite specifiers to URLs is required to get it to run in the browser. The polymer-cli's development server `polymer serve` automatically handles this transform.
+### Listeners
```
aomarks: Is this section necessary? Doesn't strike me as the kind of thing 99% of users would need to see in the docs, they would just assume the slider responds to user events.

packages/slider/README.md
```diff
@@ -1,64 +1,124 @@
...
+| `pin`     | `bookean` | `false` | Shows the thumb pin on a discrete slider.
+| `markers` | `bookean` | `false` | Shows the marker on the slider's track of a discrete slider.
 
-  1. Run the development server and open a browser pointing to its URL:
+### Methods
 
-      ```polymer serve```
+| Name     | Description
+| -------- | -------------
+| `layout() => void` | Updates the internal size model. May be required if the slider is resized or repositioned horizontally.
```
aomarks: Can you be more prescriptive about when a user should call this, similar to what we have for `layout` in https://github.com/material-components/material-components-web-components/tree/master/packages/textfield#methods? Also, how come textfield's `layout` returns a promise but this doesn't?

packages/slider/README.md
```diff
@@ -1,64 +1,124 @@
...
+### Events
 
-## Supported Browsers
-
-The last 2 versions of all modern browsers are supported, including
-Chrome, Safari, Opera, Firefox, Edge. In addition, Internet Explorer 11 is also supported.
\ No newline at end of file
+| Event Name | Target       | Detail             | Description
+| ---------- | ------------ | ------------------ | -----------
+| `input`    | `mwc-slider` | `Slider`           | Fired when the slider is beginning to open.
```
aomarks: What does it mean for a slider to "open"?

aomarks: The detail *is* the Slider?

packages/slider/README.md
e111077: Tested it. 1.1 is valid and works as expected but 0.1 gets rounded

packages/slider/README.md
e111077: BOOKEAN

packages/slider/README.md
aomarks: Do you think that's a bug, or is there some reason it makes sense to round 0-1 but not 1+?

e111077: seems to be intentional:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/foundation.ts#L229

packages/slider/README.md
e111077: textfield's awaits update complete. Perhaps I can pull that out of that function and make it sync.

packages/slider/README.md
e111077: OMG I never updated the descriptions for this XD

e111077: yes, detail is `this`

packages/slider/README.md
e111077: changed desc to what they have on MDC docs. lmk what you think

packages/slider/README.md
e111077: added warning to docs

packages/slider/README.md
```diff
@@ -1,64 +1,111 @@
...
-      ```<mwc-slider></mwc-slider>```
+### Properties/Attributes
 
-  1. Install the Polymer CLI:
+| Name      | Type      | Default | Description
+| --------- | --------- |-------- | -----------
+| `value`   | `number`  | `0`     | Current value of the slider.
+| `min`     | `number`  | `0`     | Minimum value of the slider.
+| `max`     | `number`  | `100`   | Maximum value of the slider.
+| `step`    | `number`  | `0`     | Sets the step of the slider making it discrete. Set to 0 to make continuous.<br> **Note:** If `step` is greater than 0 but less than 1, it will be rounded up to 1.<br> **Note:** Due to float math rounding issues, we recommend you set this to an integer if you also enable the pin.
```
aomarks: Suggestion from reading the MDC Web Docs:

> When a step value is given, the slider will quantize (round to the nearest multiple) all values to match that step value, except for the minimum and maximum values, which can always be set. When 0, quantization is disabled. Throws when <0.

e111077:
update on this PR?

PR #621, opened 2019-11-09
Linear progress docs + other fixes

### LINKED ISSUE #583: Document Linear Progress

e111077:
fixes #583.

[link to rendered readme](https://github.com/material-components/material-components-web-components/blob/linear-progress-docs/packages/linear-progress/README.md)

taken from changelog:

- **BREAKING** `LinearProgress.determinate = false` removed in favor of `LinearProgres.indeterminate = false`.
- **BREAKING** `LinearProgress.buffer = 0` default value changed to `1`.
- **BREAKING:VISUAL** `mwc-linear-progress` had `--mdc-theme-secondary` applied
  to its buffer bar's background color. This custom property's name was changed
  to `--mdc-linear-progress-buffer-color`.
- `mwc-linear-progress` can now have its pin and markers added and changed
  dynamically.


packages/linear-progress/README.md
```diff
@@ -1,64 +1,134 @@
-# mwc-linear-progress
+# `<mwc-linear-progress>` [![Published on npm](https://img.shields.io/npm/v/@material/mwc-linear-progress.svg)](https://www.npmjs.com/package/@material/mwc-linear-progress)
 
-> :warning: These components are a work in progress. They are pre-release and should be considered experimental, as they may undergo major changes before release. We are experimenting with alternate architectures and approaches with the goal of allowing us to bring the most correct and optimal implementation of Material components to the widest possible audiences. Visible progress may be slow, as this research is across teams and repositories so is not consistently reflected in commits to this codebase. :warning:
+> IMPORTANT: The Material Web Components are a work in progress and subject to
+> major changes until 1.0 release.
 
-A [Material Components](https://material.io/components/) linear-progress implementation using [Web Components](https://www.webcomponents.org/introduction)
+Progress indicators express an unspecified wait time or display the length of a process.
```
aomarks: Why is this reversed from the MDC description? Prefer "Progress indicators display the length of a process or express an unspecified wait time."

packages/linear-progress/README.md
```diff
@@ -1,64 +1,134 @@
...
+<img src="images/styled.gif" height="45px">
 
-  1. Run the development server and open a browser pointing to its URL:
+```html
+<style>
+  mwc-linear-progress {
+    --mdc-theme-primary: red;
+    --mdc-linear-progress-buffer-color: orange;
+    /* Note: all that was changed from default was "fill='orange'" */
+    --mdc-linear-progress-buffering-dots-image:
```
aomarks: Not needed for this PR but is this really the only way to change the color of the dots? Could we use a CSS Custom Property in the SVG instead? Does ShadyCSS support that?

(Also let's check if the new linear progress design still has the dots before doing anything.)

packages/linear-progress/README.md
e111077: lifted directly from here:
https://material.io/components/progress-indicators/#

packages/linear-progress/README.md
e111077: It's a background-image so there is no dom we can edit >.<

What they do in the SASS is that they use SASS to convert the color to url format, and then they just inject it into the url string. The problem is that, with CSS CPs, we cannot do:

```
background-image: url('.... color=var(--url-encoded-color, '%23FF00FF') ...');
```

e111077: alternatively, they can have actual SVG files with different colors and just change the `--mdc-linear-progress-buffering-dots-image` to point to the new file

PR #627, opened 2019-11-13
Use default slot for mwc-icon-button #557

### LINKED ISSUE #557: Make named slot in <mwc-icon-button> the default
**I'm submitting a:**

- [ ] bug report
- [x] feature request

**What OS are you using?**
macOS 10.14.6

All

**Current behavior:**
Currently the slot for using an SVG image is a named slot (`slot="icon"`).

**Expected behavior:**
This is the only slot, so it could be the default.

**Other information:**
We have a library of SVGs that I add with `lit-html`, as in `${logoutIcon}`. Adding a slot attribute to each SVG is a bother, and might confuse their use in other situations where the slot might be a different name. Since the SVG is a named slot, I have wrap each in a span, like `<mwc-icon-button><span slot="icon">${logoutIcon}</span></mwc-icon-button>`

asyncLiz:
Fixes #557


asyncLiz:
So to get around empty text nodes overriding the default slot's content, we're stacking the `<i>` on top of the slot. Since only either one or the other will be used, empty text nodes (for example, newlines) that use the default slot will not override the fallback `<i>`.

```html
<mwc-icon-button icon="close">
  <!-- I still show a close icon properly -->
</mwc-icon-button>
```

There are a couple ways to stack it inside the button's content area. We could use a wrapper element that is relatively positioned.

```html
<button>
  <div style="position: relative; width: 100%; height: 100%;">
    <i class="material-icons" style="position: absolute;">${this.icon}</i>
    <slot></slot>
  </div>
</button>
```

Or we could make the button relative and position the `<i>` into the right spot with the button's padding.

```html

<button style="position: relative;">
  <i class="material-icons" style="position: absolute; top: 0; padding-top: inherit;">${this.icon}</i>
  <slot></slot>
</button>
```

The wrapper seems cleaner and less hacky to me, but it's an extra DOM node to deal with that isn't strictly needed.

Any thoughts or alternative suggestions?

asyncLiz:
This should be the final result, pure CSS and passes IE/Edge alignment tests

PR #647, opened 2019-11-19
Add GitHub Action to run MWC unit tests

aomarks:
For now only runs Chrome headless and Firefox headless. In a follow up, we can add Sauce integration.


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-node@v1
+      with:
+          node-version: 10
```
azakus: Can we make this the 12 LTS, or "latest lts"?

aomarks: Pending feature request at https://github.com/actions/setup-node/issues/26 -- for now I just set 12.

aomarks:
Also made Chrome/Firefox run concurrently instead of in separate steps.

aomarks:
> Also made Chrome/Firefox run concurrently instead of in separate steps.

Which dropped time from 3m2s to 1m57s

.github/workflows/unit-tests.yml
azakus: SGTM

PR #667, opened 2019-11-29
change form-element to work with native-form

lotstar87:
sorry, I have poor english and typescript skills.

I've changed form-element to work with form element(html form element).

now, we can submit form without additional code for making form-data.

see this https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example

### DIFF
--- a/demos/index.js
+++ b/demos/index.js
@@ -14,8 +14,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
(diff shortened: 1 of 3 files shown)


lotstar87:
@googlebot I signed it!

lotstar87:
#667

e111077:
Hello, and thank you very much for the contribution! I appreciate your implementation of form association. Unfortunately, this API only works in Chrome and there are only public signals from Safari but no implementation.

Due to the fact that it is difficult to polyfill form-association, we may instead implement the form-participation API because it is easier and lighter to polyfill.

You can follow the progress on #289

PR #680, opened 2019-12-04
Textfield validation outlined

### LINKED ISSUE #660: [mwc-textfield] outlined causes setCustomValidity test to fail

**I'm submitting a:**

- [ x] bug report
- [ ] feature request

**What OS are you using?**
OSX 10.14.6

Chrome/Firefox

**Current behavior:**
If you use an outlined textfield with an initial value the component will crash.

**Expected behavior:**
You should be able to use an outlined textfield with an initial value and not have it crash.

**Steps to reproduce:**
Update the test suite for mdc-textfield so the basic fragment appears like this:
```js
const basic = html`
  <mwc-textfield outlined .value="anything"></mwc-textfield>
`;
```

**Other information:**
You'll see this error:
this.mdcFoundation is undefined
      _checkValidity@http://localhost:9876node_modules/@material/mwc-textfield/mwc-textfield-base.js:301:7
      get validity@http://localhost:9876node_modules/@material/mwc-textfield/mwc-textfield-base.js:99:10
      @http://localhost:9876test/lib/packages/textfield/src/test/mwc-textfield.test.js:115:24

This is because firstUpdated awaits the child outlined component which causes a delay in mdcFoundation being created. So by the time updated is called, mdcFoundation is undefined, but the change in value calls this.mdcFoundation.setValue. You should at least check if mdcFoundation is defined in updated and better yet, if its not, delay setting the value until after updateComplete.

e111077:
fixes #660.

Adds outlined validation tests, adds warning in readme, fixes tpyos.

### DIFF
--- a/packages/textfield/README.md
+++ b/packages/textfield/README.md
@@ -194,7 +194,7 @@ type ValidityTransform = (value: string, nativeValidity: ValidityState) => Parti
 | -------- | -------------
 | `checkValidity() => boolean`   | Returns `true` if the textfield passes validity checks. Returns `false` and fires an [`invalid`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event) event on the textfield otherwise.
 | `reportValidity() => boolean`   | Runs `checkValidity()` method, and if it returns false, then ir reports to the user that the input is invalid.
-| `setCustomValidity(message:string) => void`   | Sets a custom validity message (also overwrites `validationMessage`). If this message is not the empty string, then the element is suffering froma  custom validity error and does not validate.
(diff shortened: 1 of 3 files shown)


packages/textfield/README.md
```diff
@@ -244,6 +244,9 @@ Additionally, it implements more features such as:
 
 By default, `<mwc-textfield>` will report validation on `blur`.
 
+**NOTE:** Every property or function that checks validity must await
```
aomarks: Maybe move this note directly into the description for `checkValidity` in the method table above, since it might be easy to miss it otherwise. Also clarify if you need to do that *every* time you call `checkValidity` or just the first time.

e111077:
breakage was due to this:

https://stackblitz.com/edit/lit-element-ysqhf2?file=src%2Findex.js

I over-rid `_getUpdateComplete` to fix this.

e111077:
R4G for the sake of running internal affected tests

PR #699, opened 2019-12-13
Dialog show and close

e111077:
Added `show` and `close` methods to dialog because it is closer to `HTMLDialogElement` and because relying only on property runs into the following issue:

```ts
class MyElement extends LitElement {
  @property() isDialogOpen = true;

  render() {
    return html`
      <mwc-dialog .open=${this.isDialogOpen}></mwc-dialog>
    `;
  }
}
```

When user closes dialog with interaction, `isDialogOpen` will remain true while `mwc-dialog` remains false. If they want to open it again, they must call `this.requestUpdate('isDialogOpen')`.

### DIFF
--- a/packages/dialog/README.md
+++ b/packages/dialog/README.md
@@ -232,7 +232,10 @@ npm install @material/mwc-dialog
 | -------- | -------------
 | `forceLayout() => void` | Forces dialog to relayout (animation frame time). May be required if dialog size is incorrect or if stacked layout has not been triggered correctly.
 | `focus() => void` | Focuses on the initial focus element if defined (see [focus section](#focus)).
-| `blur() => void` | Blurs the active element.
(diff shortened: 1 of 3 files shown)


packages/dialog/README.md
```diff
@@ -232,7 +232,10 @@ npm install @material/mwc-dialog
 | -------- | -------------
 | `forceLayout() => void` | Forces dialog to relayout (animation frame time). May be required if dialog size is incorrect or if stacked layout has not been triggered correctly.
 | `focus() => void` | Focuses on the initial focus element if defined (see [focus section](#focus)).
-| `blur() => void` | Blurs the active element.
+| `blur() => void`  | Blurs the active element.
```
aomarks: This is duplicated

PR #703, opened 2019-12-14
Fix typo in mwc-snackbar README

yujiri8:


yujiri8:
@googlebot I signed it!

PR #706, opened 2019-12-16
update changelog for release

e111077:


e111077:
copybara seems to be unresponsive to the lgtm internally. Think it's perhaps because there are no affected tests?

PR #713, opened 2019-12-19
remove es6 from demos

### LINKED ISSUE #700: MWC-tabs do not even show up in IE 11

**I'm submitting a:**

- [x] bug report
- [ ] feature request

**What OS are you using?**

Windows 10 Professional, build 18362

Internet Explorer 11 

**Current behavior:**
When opening the tabs demo in IE11, the tabs do not even appear on the page

**Expected behavior:**
Tabs should show up, be clickable, have indicators, the whole nine yards

**Steps to reproduce:**

Open the MWC-tabs demo in IE11

**Other information:**
It's my understanding that the MWC components are supposed to at least be functional in IE11. The tabs are far from functional.  If you're not actually supporting IE11, then nevermind.

e111077:
fixes #700

checked that ie will now show the tabs. There are some bugs though. We should work on adding scuba tests for those failing cases

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -159,14 +159,14 @@ limitations under the License.
 
   <script>
     document.querySelectorAll('style[data-pre]').forEach(function(e) {
-      const preId = e.dataset.pre;
-      const pre = document.querySelector('#' + preId);
+      var preId = e.dataset.pre;
+      var pre = document.querySelector('#' + preId);
       if (pre) {
         pre.innerText = e.innerText;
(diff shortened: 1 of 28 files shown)


e111077:
not yet. there's a bug open-wc/open-wc#758

LarsDenBakker:
If you make these module scripts, they will be compiled. For regular scripts we need to figure out what users are expecting here with regards to polyfills, babel helpers etc.

PR #728, opened 2019-12-27
mwc-layout-grid

jurajbak:
Created new mwc-layout-grid component. It's not on your list of components, but it seems to me as a common thing which should not be missing in a library.

Btw. do you have any roadmap for material-components-web-components? Or what are the plans for the future?


jurajbak:
@googlebot I signed it!

e111077:
Thank you for the PR, but we have decided that we are only supporting components from MDC. Though, please feel empowered to release this component on your own

jurajbak:
It's already based on MDC. https://github.com/material-components/material-components-web/tree/master/packages/mdc-layout-grid

asyncLiz:
Thanks for pointing that out! At the moment we are focused on implementing the components in our README roadmap first, but I've created #844 to track this one.

This one is lower priority since there are a lot of great CSS grid systems out there already that you can use in your apps.

PR #771, opened 2020-01-15
Fix SASS

e111077:
mdc sass was migrated to modules. This needed an update to sass as well as sass-modularization of textfield, textarea, and top-app-bar.

internal investigation by sass team ongoing to see what's causing the issues in these elements


e111077:
kicking copybara to see if build system needs fixing

PR #846, opened 2020-02-06
feat(textfield): add inputmode

### LINKED ISSUE #834: mwc-textfield Needs to support inputmode
**I'm submitting a:**

- [x ] bug report
- [ ] feature request

**What OS are you using?**
Windows

Chrome

**Current behavior:**
Adding attribute `inputmode` does not cascade down to actual `input`

**Expected behavior:**
Expect the `inputmode` attribute to reflect on actual `input`

asyncLiz:
Fixes #834

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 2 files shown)


asyncLiz:
@e111077 one potential problem: `inputMode` is already defined on `HTMLElement` with a type of string, but lit-analyzer is complaining because it wants it to be more specific. If I change the type to be more specific, then TS fails to build because it is not assignable to string on the base class.

Any suggestions? Could try to bypass lit-analyzer and keep it open as string, suppress the TS problem with `@ts-ignore`, or switch to a different property like `inputmode`. Thoughts?

e111077:
I think it's simple enough to make a type for it and export it just like `input[type]`

PR #874, opened 2020-02-17
Apply select ink color

### LINKED ISSUE #869: css variable --mdc-select-ink-color not affecting mwc-select

**I'm submitting a:**

- [ x] bug report
- [ ] feature request

**What OS are you using?**
Chrome on OSX

**Current behavior:**

The CSS variable `--mdc-select-ink-color` is supposed to control the text color of the selected value in mwc-select, but it doesn't.

**Steps to reproduce:**
https://jsbin.com/jifipinini/edit?html,console,output

e111077:
fixes #869

Actually applies `--mdc-select-ink-color`

### DIFF
--- a/packages/select/src/mwc-select.scss
+++ b/packages/select/src/mwc-select.scss
@@ -114,16 +114,14 @@ mwc-notched-outline {
       }
     }
 
-    &:not(.mdc-select--invalid){
-      &:not(.mdc-select--focused) {
-        .mdc-floating-label, .mdc-floating-label::after {
-          color: var(--mdc-select-label-ink-color, #{vars.$label-color});
-        }
+    &:not(.mdc-select--invalid):not(.mdc-select--focused) {
+      .mdc-floating-label, .mdc-floating-label::after {
+        color: var(--mdc-select-label-ink-color, #{vars.$label-color});
+      }
 
-        .mdc-select__dropdown-icon {
-          color: var(--mdc-select-dropdown-icon-color, vars.$dropdown-color);
-          opacity: var(--mdc-select-dropdown-icon-opacity, vars.$dropdown-opacity);
-        }
+      .mdc-select__dropdown-icon {
+        color: var(--mdc-select-dropdown-icon-color, vars.$dropdown-color);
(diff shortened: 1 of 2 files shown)


packages/select/src/mwc-select.scss
```diff
@@ -79,11 +79,15 @@ mwc-notched-outline {
 
 :host(:not([disabled])) {
   .mdc-select {
+    .mdc-select__selected-text {
+      color: var(--mdc-select-ink-color, rgba(0,0,0,.87));
```
asyncLiz: We should probably prefer using the mixin:

```scss
.mdc-select {
  @include mixins.ink-color((
    varname: --mdc-select-ink-color,
    fallback: vars.$ink-color
  ));
}
```

packages/select/src/mwc-select.scss
```diff
@@ -79,11 +79,15 @@ mwc-notched-outline {
...
+    }
+
     &:not(.mdc-select--outlined) {
       @include mixins.container-fill-color(var(--mdc-select-fill-color, #{vars.$fill-color}));
 
       &:not(.mdc-select--invalid) .mdc-select__selected-text {
-        border-bottom-color: var(--mdc-select-idle-line-color, rgb(0, 0, 0, 0.42));
+        border-bottom-color: var(--mdc-select-idle-line-color, rgba(0, 0, 0, 0.42));
```
asyncLiz: Can we update the fallback values to correspond to their variable defaults? For example, this should be `vars.$bottom-line-idle-color`.

Additionally, there are quite a few places where we should be using mixins like above. Notably, the line ripple is no longer a border on the selected text element, but a `::before` element on `mdc-line-ripple`. Using the mixin `bottom-line-color()` would avoid this type of breaking change.

I'd also encourage the use of the [`theme.prop()` syntax](https://github.com/material-components/material-components-web/tree/master/packages/mdc-theme#mdc-theme-prop-with-css-custom-properties) for vars. It will add a fallback for browsers that don't support custom properties.

e111077: Converted the immediate ones around these changes. I believe the others would be better suited for another PR. I'll file an issue RN

PR #937, opened 2020-02-28
chore: move lint and format checks earlier in CI

e111077:
Make it take less time to get to the quick, dumb checks


e111077:
ugh moved lint back because I forgot it needs the actual TS

e111077:
adding label to test a copybara change

PR #951, opened 2020-03-03
feat(select): select can now select via value prop

### LINKED ISSUE #875: mwc-select can select via `value`
like native `select` make `mwc-select` select by setting `value`

e111077:
fixes #875

possible large performance slowdown on first render in the following case:

```html
<mwc-select value="nonempty">
  <div>Some assignedNode that is not a selectable item</div>
</mwc-select>
```

This slowdown was introduced because in the shady dom case, if you init with value, by first render callback you need your list items to be ready to select with `.value`. To prevent this wait on newer browsers, I limited it to the case where `value` was set, there are assigned nodes, but mwc-select does not know of any list-items. I felt that this would be a highly uncommon case as it is documented that setting `value` before the child is attached would not work as intended.


e111077:
copybaraing for tap run

PR #977, opened 2020-03-06
perf(list): benchmarks for list and list-item

e111077:
related #946

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "powerwash:ts": "rm packages/*/*.js & rm packages/*/*.js.map & rm packages/*/*.d.ts & rm packages/*/.tsbuildinfo",
     "powerwash:styling": "rm packages/*/src/*-css.ts",
     "powerwash:bench": "rm -rf test/lib & rm test/.tsbuildinfo*",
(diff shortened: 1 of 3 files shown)


e111077:
no need for copybara on this one

package.json
```diff
@@ -33,7 +33,7 @@
     "powerwash:ts": "rm packages/*/*.js & rm packages/*/*.js.map & rm packages/*/*.d.ts & rm packages/*/.tsbuildinfo",
     "powerwash:styling": "rm packages/*/src/*-css.ts",
     "powerwash:bench": "rm -rf test/lib & rm test/.tsbuildinfo*",
-    "powerwash": "npm run powerwash:ts & npm run powerwash:styling * npm run powerwash:bench && rm -rf node_modules",
+    "powerwash": "npm run powerwash:ts & npm run powerwash:styling & npm run powerwash:bench && rm -rf node_modules",
```
dfreedm: Do you mean for these to run concurrently?

e111077: reason is if one of them already doesn't exist e.g. ts files are not there, it'll throw something like file or directory not found and then not run the other ones

PR #983, opened 2020-03-10
[mwc-button] fix: Punctuation shows up as char reference.

smashah:
This change prevents "Let's Chat" turning into "Let &_#_39;s chat" (without the _s).

This problem specifically shows up when rendering within Storybook.

### DIFF
--- a/packages/button/src/mwc-button-base.ts
+++ b/packages/button/src/mwc-button-base.ts
@@ -77,6 +77,7 @@ export class ButtonBase extends LitElement {
       'mdc-button--outlined': this.outlined,
       'mdc-button--dense': this.dense,
     };
+    this.label = this.label.replace(/(&#(\d+);)/g, (_match, _capture, charCode) => String.fromCharCode(charCode));
     return html`
       <button
           id="button"


smashah:
Reference:

image

Expected (+ result with fix)
image

asyncLiz:
Can you provide a reproduction? This looks like it's working correctly both with the label attribute and character code for the slot.

https://jsbin.com/cukahowuva/1/edit?html,output

smashah:
@asyncLiz The only situation I can repro this issue in is in storybook (so far):

```bash
> git clone https://github.com/smashah/whatsapp-button
> cd whatsapp-button
> npm i
> npm run storybook
```

asyncLiz:
This looks like a bug with storybook, not MWC. If I launch storybook and create a new `<whatsapp-button>` element manually, it has the correct label. I'd file a bug report there!

smashah:
Thanks for your time @asyncLiz 

The issue is fixed in storybook with this: https://github.com/storybookjs/storybook/issues/4445

However, this change may still be useful as last resort? Maybe with a boolean attribute `prevent-escape`? Or close the PR. Your call.

Thanks again for your time.

asyncLiz:
I'm going to go ahead and close this, since we don't want to add specific logic to work around rendering issues with other frameworks and libraries.

PR #1002, opened 2020-03-12
fix(dialog): demo buttons open appropriate dialogs

### LINKED ISSUE #1000: Demo: Dialog buttons all open the same dialog
**Describe the bug**
On the demo page for Dialog, all of the buttons only open the same dialog "Initial Focus"

**To Reproduce**
1. Go to https://mwc-demos.glitch.me/demos/dialog.html
2. Click on any of the buttons
3. See "Initial Focus" dialog

**Expected behavior**
I think this is obvious. But nevertheless: It should open the respective dialog related to that button.

**Screenshots**
Peek 2020-03-12 16-32

 - OS: elementary OS 5.1.2 Hera
 - Browser: **Firefox**
 - Version: Dev 74.0 (64-bit)
 - Also reproduced on **Chrome**

e111077:
fixes #1000

### DIFF
--- a/demos/dialog.html
+++ b/demos/dialog.html
@@ -166,10 +166,16 @@ limitations under the License.
       var button = buttons[i];
       var buttonNum = button.dataset.num;
 
-      button.addEventListener('click', function() {
-        var dialog = document.body.querySelector('#dialog' + buttonNum);
-        dialog.open = true;
-      });
+      function listenerFactory (numButton) {
+        return function() {
+          var dialog = document.body.querySelector('#dialog' + numButton);
+          dialog.open = true;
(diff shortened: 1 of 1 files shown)


TeNNoX:
Just for the fun of smart-assing :nerd_face: - The a simpler fix could've been to use `let`, which declares a variable in block scope:  
https://codepen.io/tennox/pen/gOperQb?editors=0010

See mdn.io/let-statement

e111077:
there was a PR to remove ES6 at the time as es-dev-server was not transforming inline scripts. It does now, but I have to add type="module", but whatever /shrug

PR #1029, opened 2020-03-23
demo(slider): Fix header and glitch link

e111077:


e111077:
I don't think I'll release this one. So minor an issue

PR #1030, opened 2020-03-23
chore: add Liz to autobump CLs

e111077:
I'd love to put our rotations email on here instead but not sure how


asyncLiz:
Typically for MDC, we don't assign a reviewer for PRs unless it's a GitHub-only change. We could probably follow that same convention, just let the auto-reviewer happen internally.

e111077:
makes sense, but regardless this is a github-only change

PR #1038, opened 2020-03-25
docs(select): fix broken table about fired events

Teapot-418:
While viewing the select-package on npm, I ran over the broken table. Unfortunately, the used notation is not supported on npm.
I updated the README to use the same notation as in other README-files.


Teapot-418:
@googlebot I signed it!

PR #1033, opened 2020-03-24
Update README.md

davie-robertson:
fixed the formatting for Events

PR #1039, opened 2020-03-25
docs(select): fix broken table about fired events

Teapot-418:
While reading the README of select on npm I ran into a buggy table. Unfortunately, npm doesn't support the notation, which was in use, therefore I changed it to the same notation used in the other README-files.


asyncLiz:
Thanks for the PR! We actually already have an outstanding one for this issue though: https://github.com/material-components/material-components-web-components/pull/1033

PR #1075, opened 2020-04-03
perf(menu): benchmarking

e111077:
basic benchmarking for menu. Would be great to have a nice wiz way to get rid of the `@property` on anchor since it's just passing it down to another element without any change and forcibly causing a rerender on boot. Perhaps `asyncQuery`?


e111077:
ping

PR #1081, opened 2020-04-06
fix(slider): min and max update UI correctly

### LINKED ISSUE #1079: Slider min/max is not respected
**Describe the bug**
When setting min/max on a slider, the bounds are not respected. Dragging the thumb sets values from 0-100 regardless of min/max

**To Reproduce**
```js
render() {
  return html`
    <mwc-slider .min="2000" .max="2050" pin step="1" .value="2020"></mwc-slider>
 `;
}
```

https://webcomponents.dev/edit/H9vYUiJDCU1ilo7ELKE1

**Expected behavior**
Values out of bounds should not be set. Dragging the thumb should move the slider between bounds.

 - OS: mac
 - Browser: FF
 - Version: 75 series

e111077:
fixes #1079

### DIFF
--- a/packages/slider/src/mwc-slider-base.ts
+++ b/packages/slider/src/mwc-slider-base.ts
@@ -41,21 +41,40 @@ export class SliderBase extends FormElement {
 
   @property({type: Number})
   @observer(function(this: SliderBase, value: number) {
-    this.mdcFoundation.setValue(value);
-  })
-  value = 0;
+    // if setting both min and max at the same time, foundation values may be
+    // dirty, so must set both so they can be set correctly internally
+    try {
+      this.mdcFoundation.setMax(this.max);
+    } catch (e) { /**/ }
 
-  @property({type: Number})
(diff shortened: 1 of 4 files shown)


packages/slider/src/mwc-slider-base.ts
```diff
@@ -153,6 +157,27 @@ export class SliderBase extends FormElement {
     }
   }
 
+  updated(changed: PropertyValues) {
+    const minChanged = changed.has('min');
+    const maxChanged = changed.has('max');
+
+    if (minChanged && maxChanged) {
+      if (this.max < this.mdcFoundation.getMin()) {
```
rowanc1, line 165: I am getting some weird behavior when my min is greater than 100 when both min/max are set. The max gets overwritten.

bennypowers:
At the risk of being a nudnik, is there an indication of when this will be released?

e111077:
We should be discussing a release tomorrow so likely EoW

PR #1117, opened 2020-04-13
mdcFoundation sets value when value is not undefined

### LINKED ISSUE #1116: [mwc-textfield] not robust to undefined value
**Describe the bug**
`mwc-textfield` does not like when its value is set to undefined. 

This happens for instance when we are using the component within a Polymer component, with bindings such as : 

```html
<mwc-textfield value="[[boundTo.value]]"><mwc-textfield>
```

This cause : 
- label not to float again when a new value is set
- throw "Uncaught (in promise) TypeError: Cannot read property 'length' of undefined"

Situation is similar with `mwc-textarea`;

**To Reproduce**
https://codepen.io/christophe-g/pen/ZEbGMwp

**Expected behavior**
The component should be able to handle such case

**Screenshots**
image

christophe-g:
This should fix #1116.

### DIFF
--- a/packages/textfield/src/mwc-textfield-base.ts
+++ b/packages/textfield/src/mwc-textfield-base.ts
@@ -286,9 +286,8 @@ export abstract class TextFieldBase extends FormElement {
       this.createFoundation();
     }
 
-    if (changedProperties.has('value') &&
-        this.value !== undefined) {
-      this.mdcFoundation.setValue(this.value);
+    if (changedProperties.has('value')) {
+      this.mdcFoundation.setValue(this.value === undefined ? '' : this.value);
     }
   }
 
--- a/packages/textfield/src/mwc-textfield-base.ts
+++ b/packages/textfield/src/mwc-textfield-base.ts
@@ -287,7 +287,7 @@ export abstract class TextFieldBase extends FormElement {
     }
 
     if (changedProperties.has('value') &&
-        changedProperties.get('value') !== undefined) {
+        this.value !== undefined) {
       this.mdcFoundation.setValue(this.value);
     }
   }


packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -287,7 +287,7 @@ export abstract class TextFieldBase extends FormElement {
     }
 
     if (changedProperties.has('value') &&
-        changedProperties.get('value') !== undefined) {
```
asyncLiz, line 290: This was added to fix https://github.com/material-components/material-components-web-components/issues/515. I'm not sure what this change is hoping to achieve by removing the functionality.

This goes from "if the value _was previously_ not undefined" to "if the value _is currently_ not undefined", which isn't what the issue was solving.

christophe-g: If I understand correctly, the fix was about to call Foundation `setValue`  when value changes. This does not change. 

#515 and #1116 have a similar root cause which is the component's inability to re-evaluate positioning logic when value change. 

Then problem of `changedProperties.get('value') !== undefined)` is that when you have an `undefined` value at some stage, the positioning logic fail again to be called  next time the value changes (see https://codepen.io/christophe-g/pen/ZEbGMwp)

packages/textfield/src/mwc-textfield-base.ts
```diff
@@ -286,9 +286,8 @@ export abstract class TextFieldBase extends FormElement {
       this.createFoundation();
     }
 
-    if (changedProperties.has('value') &&
-        this.value !== undefined) {
-      this.mdcFoundation.setValue(this.value);
+    if (changedProperties.has('value')) {
+      this.mdcFoundation.setValue(this.value === undefined ? '' : this.value);
```
christophe-g, line 290: this treats undefined values as empty strings and avoids 
image

asyncLiz: This is actually the exact type of behavior I want us to avoid doing. This adds logic to support setting a property to an invalid type. According to typescript, `this.value` should never be `undefined` because it must be a strict String type.

We don't have any extra logic to handle `this.value = new Date()`, even though it could be a valid use case for `type="date"` inputs. It'd be an ever-growing maintenance cost to continually add logic for all the different types a property could be set to. If there's a valid use case for the property to be undefined, it should be typed as `string | undefined`.

If we're wanting to achieve native `<input>` parity, which uses a strict `string` for it's value (it can't be undefined), then we should be coercing all values to a string. Setting `textfield.value = undefined` would be the same as `textfield.value = String(undefined)`.

e111077:
closed in favor of #1313

PR #1244, opened 2020-04-30
mwc-select: remove aria required attribute

### LINKED ISSUE #1237: mwc-select fails chrome a11y check for aria attributes
**Describe the bug**

Chrome lighthouse a11y audit fails on select with the following error:

`[aria-*] attributes do not match their roles`

**To Reproduce**
1. Open select demo
2. run accessibility audit in chrome

**Expected behavior**

no audit failure

43081j:
here you go @e111077 it fixes #1237 

a button role doesn't have a "required" attribute

### DIFF
--- a/packages/select/src/mwc-select-base.ts
+++ b/packages/select/src/mwc-select-base.ts
@@ -259,7 +259,6 @@ export abstract class SelectBase extends FormElement {
               aria-invalid=${!this.isUiValid}
               aria-haspopup="listbox"
               aria-labelledby="label"
-              aria-required=${this.required}
               aria-describedby=${ifDefined(describedby)}
               @click=${this.onClick}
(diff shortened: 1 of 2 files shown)


e111077:
Thanks for the PR, but I spoke with the Chrome a11y standards team and they think that the button role is a bit overloaded and aira-required should probably live on role="button" in the specific case of a select. We're moving to make a spec change proposal

e111077:
On a side note I've contacted the Apple Mac Voice Over team in regards to their tool not announcing "required" when set on role="listbox"

43081j:
Awesome, did feel a bit wrong getting rid of it haha!

i noticed the tests are old style too without the factory pattern we started using. So might be worth refactoring those some time.

PR #1343, opened 2020-05-21
updating chip prototype branch to the master and recent API/FW changes

gullerya:
Chip MWC functionality is not yet delivered, but desired and I'd like to at least get/keep? it closest to the things going on on master so that the continuation of work on this part will be smooth.
This PR aligns the chip folder (functionality) with the rest of the mwc with regard to dependencies, demoing code self is made working again so that the actual chip components may be seen in the branch in the demo page.


gullerya:
squashing and coming back :)

asyncLiz:
Thanks for the interest! However, I will go ahead and give you a heads up that the reason we're holding off on chips is because they're going through quite a few breaking changes to address some a11y concerns.

We love the enthusiasm, but I hope we can set a few expectations that the chips still aren't quite ready to be implemented until they're done with those changes in the upstream library.

PR #1348, opened 2020-05-23
fix typo in nomodule attribute

lemming:


lemming:
@googlebot I signed it!

Issue #1215, opened 2020-04-24
Add missing icon/helper-text foundation features to textfield


asyncLiz:
**Is your feature request related to a problem? Please describe.**
`<mwc-textfield>` does not use the MDCHelperTextFoundation or MDCIconFoundation. Both of these classes offer features that are missing and should be added:

Helper Text
- announcement to screen readers

Icon
- Interactivity (focus, click events, disabled)
- Aria labels

**Describe the solution you'd like**
Helper text should toggle `role="alert"` when appropriate to announce to screen readers.

Add the following properties for icons:

- `iconAriaLabel` *
- `iconInteractive` *
- `trailingIconAriaLabel`
- `trailingIconInteractive`

* Discuss with design, leading icons might only be allowed to be decorative

Dispatch the following events when an interactive icon is clicked:

- `@iconclick`
- `@trailingiconclick`

**Describe alternatives you've considered**
We could instead implement the foundations for each of these, but this would conflict with future internal tooling which would need to remove the foundations anyway.

**Additional context**
MDC foundation links:
- https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/helper-text/foundation.ts
- https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/icon/foundation.ts

DrewLandgraveCbsi:
Ran across this today. Any idea on when this could be addressed?

DrewLandgraveCbsi:
Any update on the interactivity portion of this request?

christophe-g:
In light of the approach  proposed in #893 by @dfreedm , is the integration of MDCHelperTextFoundation still the preferred approach or is it not simpler to handle the logic of `aria-hidden`, `role=alert`, `aria-describedby` accessibility attributes directly in the render function?

I am asking because I tried to add and integrate a MDCHelperTextFoundation adapter to textfield (because I need more accessible fields like ... now) and am under the impression that hooking adapters into the component makes things unnecessary bloated. 

Lit-element provides us with flexible and expressive ways to make UI depend on state. As a developer, I have the impression I am loosing that under the current implementation, with no obvious benefit (but I might miss smth).

asyncLiz:
@christophe-g this issue plans to add these features directly to the component _without_ the use of MDCHelperTextFoundation or MDCIconFoundation:

> **Describe alternatives you've considered**
We could instead implement the foundations for each of these, but this would conflict with future internal tooling which would need to remove the foundations anyway.

[...]

Issue #602, opened 2019-10-31
Research extensibility


govis:
We tried extending <code>TabBarBase</code> to activate tab by its' id as opposed to index. To do this one needs access to <code>\_getTabs</code> method, but it's private and call <code>getIndexOfTabById</code>, which is buried in <code>mdcFoundation.adapter_</code> and is not accessible either

kr05:
Most of these components are hard to extend. From personal experience, I've gone through so much trouble customizing:

mwc-button
mwc-textfield
mwc-fab
mwc-drawer
mwc-radio
mwc-switch

and probably a couple more I forgot. Most of the times, I simply make the changes locally and simply make sure I remember to use the same machine to build the project.

Here's an example of a simple change I made to the css file for mwc-drawer:
```css
.mdc-drawer-app-content {
  /* Default styles */
  margin-left: 0;
  margin-right: 0;
  position: relative;

  /* Added style */
  width: 100%;
}
```
On more important client projects, I have unfortunately been forced to either accept the component for what it is, or choose an alternative.

vdegenne:
I agree, I don't remember having issues to customize mwc elements on the fly but have been using `mwc-tab-bar` in few projects now and hell I always spend so much time trying to harmonize its logic with my application. Maybe providing some user-friendly accessibility functions is a good idea ?

kr05:
For a specific example, it would be nice to be able to use actionItems other than mwc-icon-buttons.

azakus:
We're currently exploring theming in general in #526.
Are you extending just to modify styling, or do you want to modify behavior as well?

If you just need styling changes, it seems likely we just need to set all `private` API to `protected` instead.

govis:
Yes, <code>protected</code> API and cleaner way to access <code>adapter</code> methods would be nice

aomarks:
Can you give more details of what you're trying to do? It will help us understand what kind of customization users need. For example, why do you need to call `_getTabs`?

govis:
<pre><code>&lt;my-tab-bar active-tab-id="${this._tab}" @MDCTabBar:activated="${this._tabActivated}">
    &lt;mwc-tab id="tab1">Tab 1&lt;/mwc-tab>
    ${this._someCondition ? html`
        &lt;mwc-tab id="tab2">Tab 2&lt;/mwc-tab>
    ` : ''}
&lt;/my-tab-bar></code></pre>

<pre><code>class MyTabBar extends TabBarBase {
[...]

PR #1437, opened 2020-06-18
feat(textfield): add slots for prefix and suffix

davemooreuws:
Fixes one of the issues here: https://github.com/material-components/material-components-web-components/issues/1215

### DIFF
--- a/packages/textfield/src/mwc-textfield-base.ts
+++ b/packages/textfield/src/mwc-textfield-base.ts
@@ -270,9 +270,9 @@ export abstract class TextFieldBase extends FormElement {
       <label class="mdc-text-field ${classMap(classes)}">
         ${ripple}
         ${this.icon ? this.renderIcon(this.icon) : ''}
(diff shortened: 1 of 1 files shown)


davemooreuws:
@googlebot I signed it!

davemooreuws:
@googlebot I fixed it

asyncLiz:
Thanks for the PR!

Unfortunately per spec, prefixes and suffixes may only be text and should not be any other type of html content. This is the reason we decided to make them string properties instead of slots.

As discussed in https://github.com/material-components/material-components-web-components/issues/602, our plan is to make our components easily extensible for users who wish to add additional features like slot-able affixes that aren't part of the strict spec that we'd adhering to. If you wanted a slot for prefix/suffix, you can extend the textfield and override `render()` or `renderAffix()` to add the slot.

To fix #1215, we'll need to add additional properties to support the features described instead of adding slots.

PR #1479, opened 2020-06-29
Prevent mwc-list-item selectedText from including text from font liga…

WeftDigital:
…tures of slotted icons

Problem: the `selectedText` property of mwc-list-item returns the text content of all slots, including the font ligatures of slotted mwc-icons. This currently prevents us from using slotted icons with mwc-list-items inside of an mwc-select because the displayed `selectedText` will also include the font ligatures' text.

Fix: Return only the text of the default/primary slot, not that of the icon slots.

I also added "Options with Icons" to the mwc-select demo page.

### DIFF
--- a/packages/list/src/mwc-list-item-base.ts
+++ b/packages/list/src/mwc-list-item-base.ts
@@ -154,13 +154,15 @@ export class ListItemBase extends LitElement {
       ];
 
   get text() {
-    const textContent = this.textContent;
-
-    return textContent ? this.renderRoot.querySelectorAll('slot')[1].assignedNodes().reduce(
+    const primaryTextSlot = this.renderRoot.querySelector('.mdc-list-item__text')?.querySelector('slot');
+    if (!primaryTextSlot) {
+      return '';
+    }
+    return primaryTextSlot.assignedNodes().reduce(
       (totalString, partialStringElem)=>{
         totalString = totalString + partialStringElem.textContent;
         return totalString;
-      }, '').trim(): '';
+      }, '').trim();
   }
 
   render() {
--- a/packages/list/src/mwc-list-item-base.ts
+++ b/packages/list/src/mwc-list-item-base.ts
@@ -156,7 +156,11 @@ export class ListItemBase extends LitElement {
   get text() {
     const textContent = this.textContent;
 
-    return textContent ? textContent.trim() : '';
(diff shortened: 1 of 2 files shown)


WeftDigital:
> @googlebot I signed it!

e111077:
What are the use cases for this? The material spec does not allow for icons in select; confirmed with design.

WeftDigital:
> What are the use cases for this? The material spec does not allow for icons in select; confirmed with design.

Ah, okay. I should have checked. That's my bad. Although, what about icons in selectable lists in general?

https://glitch.com/edit/#!/scarlet-separate-armadillo?path=index.js%3A2%3A28

packages/list/src/mwc-list-item-base.ts
```diff
@@ -154,9 +154,15 @@ export class ListItemBase extends LitElement {
       ];
 
   get text() {
-    const textContent = this.textContent;
-
-    return textContent ? textContent.trim() : '';
+    const primaryTextSlot = this.renderRoot.querySelector('.mdc-list-item__text')?.querySelector('slot');
```
e111077, line 157: For SSR reasons, we cannot access `renderRoot` please replace with `@query`.

For Angular compatibility, we cannot use optional chaining.

packages/list/src/mwc-list-item-base.ts
```diff
@@ -154,9 +154,15 @@ export class ListItemBase extends LitElement {
...
+    if (!primaryTextSlot) {
+      return '';
+    }
+    return primaryTextSlot.assignedNodes().reduce(
```
e111077, line 161: For SSR reasons, we cannot use `assignedNodes()`; please find a way to do this with `@material/mwc-base/utils.ts`'s `findAssignedElement` function

WeftDigital:
Elliot, thanks for your response. You've convinced me. (I secretly still want to use icons within select opts, but admittedly I'm something of an icon fanatic, and if the design team thinks that's a bad practice, then I'm sure they have some thoughtful justifications for that decision.) 

I appreciate the info you shared about SSR and Angular compatibility though, as that was all news to me. Should I go ahead and close out this pull request? I'd love to find a way to contribute to this project. Are there any additional guidelines you could point me to beyond what's in the contributing guide to help me with future attempts to contribute? Thanks!

e111077:
I'll bring up updating our contributing docs with our team. Thanks for the help though!

Issue #1054, opened 2020-03-31
mwc-checkbox inside mwc-dialog has misaligned mouseover highlight


CaptainCodeman:
**Describe the bug**
The mouseover highlight for the mwc-checkbox is misaligned when it's inside an mwc-dialog

**To Reproduce**
Render the checkbox inside a dialog, e.g.:
```js
  render() {
    return html`
      <mwc-dialog heading="Edit Permission">
        <div>
          <mwc-formfield label="Admin">
            <mwc-checkbox ?checked=${this.isAdmin}></mwc-checkbox>
          </mwc-formfield>
        </div>
        <mwc-button dialogAction="ok" slot="primaryAction">ok</mwc-button>
        <mwc-button dialogAction="cancel" slot="secondaryAction"> cancel </mwc-button>
      </mwc-dialog>
    `
  }
```

**Expected behavior**
The mouseover highlight should be the same as the ripple

**Screenshots**
![Screen-Recording-2020-03-31-at-9](https://user-images.githubusercontent.com/304910/78043869-1e781780-7331-11ea-9cec-961bce6e7630.gif)

**Browser Version (please complete the following information):**
Chrome 80

**Additional context**
Checkbox highlight works normally when not in the dialog, no styling is applied to the dialog or the checkbox. Once clicked, the highlight appears correct. Same behavior regardless of whether mwc-formfield is used.

dfreedm:
Hmm, this might be fixed by the new checkbox I have in master. Will repro with that and confirm either way.

dfreedm:
Related #1032

dfreedm:
This apparently is an upstream issue as well https://github.com/material-components/material-components-web/issues/5443

Misiu:
@dfreedm in upstream issue there is a fix: https://stackblitz.com/edit/mdc-dialog-ripple-alignment-eoogdd?file=index.js
proposed by @asyncLiz here: https://github.com/material-components/material-components-web/issues/5443#issuecomment-572716990
any ideas on how to fix this here?

asyncLiz:
Fixed in M3

Issue #894, opened 2020-02-20
Deprecate and remove ripple directive


dfreedm:
Each component that has a ripple currently embeds ripple styling from MDC, most of which is duplicate styling.

To remove this duplication, we should remove usage of the ripple mixin from these components, and replace the ripple directive (which relies on this styling) with `mwc-ripple`.

Related #892

asyncLiz:
Closing as obsolete, we have `<mwc-ripple>` now

PR #1534, opened 2020-07-17
Refactor switch replace ripple directive with lazy mwc-ripple

bramkragten:
Refactor switch replace ripple directive with lazy `mwc-ripple`

Related #894
Related #1054

### DIFF
--- a/packages/switch/src/mwc-switch-base.ts
+++ b/packages/switch/src/mwc-switch-base.ts
@@ -76,11 +76,14 @@ export class SwitchBase extends FormElement {
   });
 
   protected renderRipple() {
-    if (this.shouldRenderRipple) {
-      return html`<mwc-ripple .accent="${this.checked}" .disabled="${
-          this.disabled}" unbounded></mwc-ripple>`;
-    }
-    return html``;
+    return html`${
+        this.shouldRenderRipple ? html`
+        <mwc-ripple 
+          .accent="${this.checked}" 
+          .disabled="${this.disabled}" 
+          unbounded>
+        </mwc-ripple>` :
+                                  html``}`;
   }
 
   focus() {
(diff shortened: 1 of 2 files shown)


packages/switch/src/mwc-switch-base.ts
```diff
@@ -63,29 +70,95 @@ export class SwitchBase extends FormElement {
...
-  get ripple() {
-    return this.rippleNode.ripple;
+  protected rippleHandlers: RippleHandlers = new RippleHandlers(() => {
+    this.shouldRenderRipple = true;
+    return this.ripple;
+  });
+
+  protected renderRipple() {
+    if (this.shouldRenderRipple) {
+      return html`<mwc-ripple .accent="${this.checked}" .disabled="${
```
e111077, line 80: Mind manually formatting this line so the formatter doesn't mangle the bindings?

packages/switch/src/mwc-switch-base.ts
```diff
@@ -63,29 +70,95 @@ export class SwitchBase extends FormElement {
...
   }
 
-  get ripple() {
-    return this.rippleNode.ripple;
+  protected rippleHandlers: RippleHandlers = new RippleHandlers(() => {
+    this.shouldRenderRipple = true;
+    return this.ripple;
+  });
+
+  protected renderRipple() {
```
e111077, line 78: Currently this if statement is unsupported in our internal SSR tooling. Can you please refactor this fn to be the same as in checkbox as an ugly ternary? Eventually we should support this pattern, but the tooling is under development.

e111077:
small error on my part; I actually did not put this change on the google-wide test train, so this may take a bit longer since that takes about half a day to run

PR #1535, opened 2020-07-17
Fix ripple deactivate on checkbox when releasing outside of checkbox

bramkragten:
Fix ripple deactivate on checkbox when releasing outside of checkbox

### DIFF
--- a/packages/checkbox/src/mwc-checkbox-base.ts
+++ b/packages/checkbox/src/mwc-checkbox-base.ts
@@ -96,12 +96,14 @@ export class CheckboxBase extends FormElement {
   /** @soyCompatible */
   protected renderRipple() {
     const selected = this.indeterminate || this.checked;
-    if (this.shouldRenderRipple) {
-      return html`<mwc-ripple .accent="${selected}" .disabled="${
-          this.disabled}" .unbounded="${true}"></mwc-ripple>`;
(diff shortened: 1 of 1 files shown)


packages/checkbox/src/mwc-checkbox-base.ts
```diff
@@ -167,32 +169,43 @@ export class CheckboxBase extends FormElement {
       </div>`;
   }
 
-  private _handleFocus() {
-    this.focused = true;
-    this.rippleHandlers.startFocus();
-  }
+  @eventOptions({passive: true})
```
dfreedm, line 172: This line can probably be removed, as I don't think making `mousedown` passive affects anything.

packages/checkbox/src/mwc-checkbox-base.ts
```diff
@@ -167,32 +169,43 @@ export class CheckboxBase extends FormElement {
       </div>`;
   }
 
-  private _handleFocus() {
-    this.focused = true;
```
dfreedm, line 171: Oh, looks like this is missing now. Can you add it back to `handleRippleFocus` and `handleRippleBlur`?

bramkragten: Oops...

PR #1536, opened 2020-07-17
chore(slider): Stop updating MDC slider dep.

joyzhong:
We'll remove this check once we've updated to the new M2 slider.


joyzhong:
> Can you also add a line to the CHANGELOG.md file noting that the slider is no longer updating in preparation for M2?
> 
> If you want that can be done separately, since it'll need to be imported to g3. Lemme know which you'd prefer.

Sure! I'll send out a separate change for that.

joyzhong:
Liz, can you merge this in when you get the chance? Thanks!

PR #1537, opened 2020-07-17
Refactor radio replace ripple with lazy mwc-ripple

bramkragten:
Refactor radio replace ripple with lazy mwc-ripple

And fix a bug where if you would focus a radio group for the second time while not focussing other radio groups it would not set focus on the selected element

### DIFF
--- a/packages/radio/src/mwc-radio-base.ts
+++ b/packages/radio/src/mwc-radio-base.ts
@@ -14,12 +14,15 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import {addHasRemoveClass, FormElement, HTMLElementWithRipple} from '@material/mwc-base/form-element';
+import '@material/mwc-ripple/mwc-ripple';
+
+import {Ripple} from '@material/mwc-ripple/mwc-ripple';
+import {RippleHandlers} from '@material/mwc-ripple/ripple-handlers';
+import {addHasRemoveClass, FormElement} from '@material/mwc-base/form-element';
 import {observer} from '@material/mwc-base/observer';
-import {ripple} from '@material/mwc-ripple/ripple-directive';
 import {MDCRadioAdapter} from '@material/radio/adapter';
 import MDCRadioFoundation from '@material/radio/foundation';
-import {html, property, query} from 'lit-element';
+import {html, property, query, queryAsync, internalProperty, eventOptions} from 'lit-element';
 import {SingleSelectionController} from './single-selection-controller';
 
 /**
@@ -30,7 +33,7 @@ export class RadioBase extends FormElement {
 
   @query('input') protected formElement!: HTMLInputElement;
 
-  @query('.mdc-radio__ripple') protected rippleElement!: HTMLElementWithRipple;
+  @queryAsync('mwc-ripple') ripple!: Promise<Ripple|null>;
 
   private _checked = false;
 
@@ -105,6 +108,8 @@ export class RadioBase extends FormElement {
 
   private _selectionController?: SingleSelectionController;
 
+  @internalProperty() protected shouldRenderRipple = false;
+
   connectedCallback() {
     super.connectedCallback();
     // Note that we must defer creating the selection controller until the
@@ -136,14 +141,11 @@ export class RadioBase extends FormElement {
 
   focus() {
     this.focusNative();
+    this.handleRippleFocus();
   }
 
   focusNative() {
     this.formElement.focus();
-  }
-
-  get ripple() {
-    return this.rippleElement.ripple;
   }
 
   protected createAdapter(): MDCRadioAdapter {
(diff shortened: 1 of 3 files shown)


bramkragten:
Another option for the focus issue would be introducing a blur function:

```ts
  /**
   * Remove focused set when no element of the set has focus anymore, or if overridden by mouse
   *
   * @param element Element from which selection set is derived and subsequently
   *     blurred.
   */
  async blur(element: CheckableElement) {
    const set = this.getSet(element.name);
    const currentSelected = set.selected;
    if (!currentSelected) {
      return;
    }
    // wait for update set
    await new Promise((resolve) => setTimeout(resolve, 0));
    if (set == this.focusedSet && this.focusedSet.selected == currentSelected) {
      this.focusedSet = null;
    }
  }
```

packages/radio/src/single-selection-controller.ts
```diff
@@ -213,15 +211,14 @@ export class SingleSelectionController {
...
    *     focused.
    */
-  focus(element: CheckableElement) {
+  async focus(element: CheckableElement) {
     // Only manage focus state when using keyboard
     if (this.mouseIsDown) {
       return;
     }
     const set = this.getSet(element.name);
-    const currentFocusedSet = this.focusedSet;
```
e111077, line 222: Why was this removed?

bramkragten: It is no longer needed, currently, there is a bug that if you focus an element of the set for the second time it will not move the focus to the selected item because the set that is being focused is the `focusedSet` (it is not cleared on `blur` of the set). This was probably added for when you change the value with your keyboard as the `focus` event comes before the `update` call. By awaiting a promise we check the selected value after the `update` call so we don't have to handle the value change of keyboard anymore (the focused element will be the selected element) and a second focus will also work now.

Should probably add a check to see if the element is still focused.

There are other ways to handle this, like the `blur` handler in my comment above. That will keep the `focusedSet` and will clear the `focusedSet` when you `blur` the entire set or on mouse override.

packages/radio/src/mwc-radio-base.ts
```diff
@@ -155,6 +158,19 @@ export class RadioBase extends FormElement {
...
   }
 
+  protected rippleHandlers: RippleHandlers = new RippleHandlers(() => {
+    this.shouldRenderRipple = true;
+    return this.ripple;
+  });
+
+  protected renderRipple() {
+    if (this.shouldRenderRipple) {
+      return html`<mwc-ripple .accent="${this.checked}" .disabled="${
```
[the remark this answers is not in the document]
bramkragten: Got to say I hate the formatter 😄  why don't you use prettier?

packages/radio/src/single-selection-controller.ts
e111077: I can't seem to repro your issue. Is there an issue filed or can you send me a repro case?

packages/radio/src/mwc-radio-base.ts
```diff
@@ -136,16 +142,13 @@ export class RadioBase extends FormElement {
 
   focus() {
     this.focusNative();
+    this.handleRippleFocus();
```
e111077, line 145: I assume this doesn't need to be added to a blur fn because blur is implicit in a radio group?

packages/radio/src/mwc-radio-base.ts
```diff
@@ -155,6 +158,23 @@ export class RadioBase extends FormElement {
     };
   }
 
+  protected rippleHandlers: RippleHandlers = new RippleHandlers(() => {
+    this.shouldRenderRipple = true;
+    return this.ripple;
+  });
+
+  protected renderRipple() {
+    if (this.shouldRenderRipple) {
```
e111077, line 167: Currently this if statement is unsupported in our internal SSR tooling. Can you please refactor this fn to be the same as in checkbox as an ugly ternary? Eventually we should support this pattern, but the tooling is under development.

packages/radio/src/single-selection-controller.ts
bramkragten: You can repo it here: <https://material-components.github.io/material-components-web-components/demos/formfield/>

As you can see in the GIF, the first time the first radio gets focus it is correctly setting the focus on the selected (second) radio element, but after moving the focus back to the switch and then again to the first radio it is not set to the selected but keeps the focus on the first radio element.
radio-focus

packages/radio/src/mwc-radio-base.ts
bramkragten: I just realized that this would not work in all cases and wondered why I didn't noticed that in my tests, I see now that it `mdc-radio` has its own focus styling `.mdc-radio .mdc-radio__background::before` & `.mdc-radio__native-control:focus + .mdc-radio__background::before`

Is this something that should be removed and replaced by the ripple or should we not use the ripple for focus?

bramkragten: The current styling doesn't match material, as focus is always in the accent color.
image

e111077: Hmm I'm going to have to pass this onto @dfreedm since he's been working with radio more-closely

dfreedm: I believe that styling is holdover from MDC for a CSS-only ripple. You should be able to turn it off by adding a class named `mdc-radio--upgraded` to the `mdcRoot` node

dfreedm: Hmm, radio works a bit differently than checkbox, the right classname is actually `mdc-ripple-upgraded`

bramkragten: That doesn't seem to work, but this does:
```scss
  .mdc-radio__background::before {
    content: none;
  }
```

packages/radio/src/single-selection-controller.ts
bramkragten: I changed it to the blur logic to not mess with setting focus, that should stay sync.

taylor-vann:
Fixed with PR #1684

PR #1540, opened 2020-07-17
Add form validation to dialog demo

taylor-vann:


demos/dialog/index.js
```diff
@@ -1,3 +1,5 @@
+import {LitElement, html} from 'lit-element/lit-element.js';
```
asyncLiz, line 1: I would avoid using LitElement in these demos. We want to show off pure JS that can copy/pasted anywhere, and using lit defeats that purpose.

demos/dialog/index.js
```diff
@@ -34,3 +36,61 @@ window.toggleActions.onclick = function() {
...
+            dialogInitialFocus
+            @input="${this.onInputListener}"
+            label="Name"
+            pattern="^[a-zA-Z]{3,}$"
+            placeholder="Casey"
+            required
+            type="text">
+        </mwc-textfield>
+        <mwc-button
+          dialogAction="close"
```
asyncLiz, line 83: I think the pattern we want to show off is to avoid dialogAction with form validation and to do it manually, right? I don't think it'd be necessary to disable/enable the button.

Instead, on clicking the button, you can check the text field's validity and then choose whether or not to manually close the dialog.

demos/dialog/index.js
taylor-vann: totally! i was still in lit-html mode from some pair coding with Dan.
I can do pure JS! i like pure JS.

demos/dialog/index.js
taylor-vann: I see!
Just to clarify, what do we mean by "manually close the dialog" ? Do we want to close the dialog once the contents of a textfield is valid programmatically? or simply don't let our users close the close until text field has valid input? :)

asyncLiz: I would say close the dialog rather than disable the button to prevent closing.

The common use case we get is a request for "how do I stop or cancel a dialogAction?" Usually the problem is the developer doesn't know whether or not the value is valid by the time they click the button to close the dialog, such as waiting on a server response.

```js
document.querySelector('.close-button').addEventListener('click', () => {
  const valid = checkValid(); // validate, possible asynchronous such as a server response
  if (valid) {
    // manually close the dialog
    dialog.close();
  } else {
    // don't close the dialog and let the user know there's an error
    // such as by setting the textfield's error message
    textfield.setCustomValidity('There was a problem');
    textfield.reportValidity();
  }
});
```

taylor-vann: Nice! will do! thanks for the example :) I'll update this

demos/dialog/index.html
```diff
@@ -153,6 +155,30 @@
         Secondary
       </mwc-button>
     </mwc-dialog>
+
+    <mwc-button data-num="8" raised>Form Validation</mwc-button>
+    <mwc-dialog id="dialog8" heading="Form Validation">
+      <p>We can validate user input before closing our dialog.</p>
+      <mwc-textfield
+        id="dialog8-text-field"
+        type="text"
```
asyncLiz, line 164: nit: `type="text"` is unnecessary, since that's the default type of `<mwc-textfield>`

demos/dialog/index.html
```diff
@@ -153,6 +155,30 @@
...
+        pattern="^[a-zA-Z\s]{3,64}$"
+        placeholder="Casey"
```
asyncLiz, line 166: What does this placeholder represent? Is it supposed to be a first name prompt? If so, I think "First name" would be a more appropriate placeholder rather than a sample first name.

demos/dialog/index.html
```diff
@@ -153,6 +155,30 @@
...
       </mwc-button>
     </mwc-dialog>
+
+    <mwc-button data-num="8" raised>Form Validation</mwc-button>
+    <mwc-dialog id="dialog8" heading="Form Validation">
+      <p>We can validate user input before closing our dialog.</p>
+      <mwc-textfield
+        id="dialog8-text-field"
+        type="text"
+        pattern="^[a-zA-Z\s]{3,64}$"
```
asyncLiz, line 165: The `minlength` and `maxlength` attributes would be more appropriate than trying to encode them into the pattern.

demos/dialog/index.js
```diff
@@ -34,3 +34,17 @@ window.toggleActions.onclick = function() {
   dialog.hideActions = hideAction;
   hideActionSpan.innerText = hideAction;
 };
+
+const textField = document.getElementById('dialog8-text-field');
+const primaryButton = document.getElementById('dialog8-primary-action-button');
+    
+primaryButton.addEventListener('click', (e) => {
+  // validate, possible asynchronous such as a server response
```
asyncLiz, line 42: What happens if the validation is asynchronous? You will not be able to stop the propagation of the click event because you'll be waiting for a server response. The click will continue to the next listener and the `dialogAction="close"` will be handled in the dialog's internal listener.

I do not believe you should be decorating the `slot="primaryAction"` button with `dialogAction="close"`. Instead, explicitly encourage users _not_ to do that and manually close the dialog with `dialog.close()`.

That will give users the most flexibility. This approach relies on the validation being synchronous so that you can call `stopPropagation()`. It also relies on the fact that this click listener is firing first and not the click listener within our dialog's implementation.

demos/dialog/index.html
```diff
@@ -153,6 +155,30 @@
         Secondary
       </mwc-button>
     </mwc-dialog>
+
+    <mwc-button data-num="8" raised>Form Validation</mwc-button>
+    <mwc-dialog id="dialog8" heading="Form Validation">
+      <p>We can validate user input before closing our dialog.</p>
```
asyncLiz, line 161: I would avoid first-person verbiage like "we" when tech writing. Instead, this could be written as "The dialog can validate user input before closing."

demos/dialog/index.js
```diff
@@ -34,3 +34,17 @@ window.toggleActions.onclick = function() {
   dialog.hideActions = hideAction;
   hideActionSpan.innerText = hideAction;
 };
+
+const textField = document.getElementById('dialog8-text-field');
```
asyncLiz, line 38: Use `querySelector('#id')` instead of `getElementById('id')`

demos/dialog/index.html
taylor-vann: that makes sense!

demos/dialog/index.html
taylor-vann: simpler too. i agree

demos/dialog/index.html
taylor-vann: sounds good!

demos/dialog/index.js
taylor-vann: kewl kewl

demos/dialog/index.js
taylor-vann: i see!

demos/dialog/index.html
taylor-vann: done!

demos/dialog/index.html
taylor-vann: done

demos/dialog/index.html
taylor-vann: done

demos/dialog/index.js
taylor-vann: done

demos/dialog/index.html
taylor-vann: done

demos/dialog/index.js
taylor-vann: done

demos/dialog/index.html
```diff
@@ -153,6 +155,29 @@
...
     </mwc-dialog>
+
+    <mwc-button data-num="8" raised>Form Validation</mwc-button>
+    <mwc-dialog id="dialog8" heading="Form Validation">
+      <p>This dialog can validate user input before closing.</p>
+      <mwc-textfield
+        id="dialog8-text-field"
+        minlength="3"
+        maxlength="64"
+        placeholder="First Name"
```
asyncLiz, line 166: nit: prefer sentence-style capitalization (go/material/writing/grammar-punctuation): "First name"

demos/dialog/index.html
```diff
@@ -153,6 +155,29 @@
...
+        required>
+      </mwc-textfield>
+      <mwc-button
+        id="dialog8-primary-action-button"
+        slot="primaryAction">
+        confirm
```
asyncLiz, line 172: Same nit here and for the cancel button. Even though these are Transformed to upper case, it's still a good habit to write them in sentence case: "Confirm" and "Cancel"

demos/dialog/index.html
taylor-vann: sweet! let's do it

demos/dialog/index.html
taylor-vann: yes, makes sense!

PR #1557, opened 2020-07-22
Add an example and documentation reference

### LINKED ISSUE #1514: mwc-dialog field validation
What is the recommended way to validate data entered by the user in an mwc-dialog?

Following the examples I have OK/Cancel buttons on my dialog. The OK button uses dialogAction="ok" which closes the dialog. I could listen for the closed event and validate the data there but that doesn't seem to be a good idea. The user could have entered data in a dozen fields with one of them invalid. Since my validation error would come after the dialog has been closed, the user would have to enter all data again.

Should I use a button without dialogAction="ok", validate and close the dialog myself? What is the recommended pattern?

YonatanKra:
Closes #1514

### DIFF
--- a/demos/dialog/index.html
+++ b/demos/dialog/index.html
@@ -155,6 +155,22 @@ limitations under the License.
         Secondary
       </mwc-button>
     </mwc-dialog>
+
+    <mwc-button data-num="8" raised>Validation in a Dialog</mwc-button>
+    <mwc-dialog id="dialog8" heading="Validation in a Dialog">
+      <div>
+        In this example we set validate a form inside a dialog before closing it.
+      </div>
+      <form id="aForm">
+        <input required name="firstName"/>
+        <mwc-button slot="primaryAction" onclick="validateAndClose()">
+          Validate and Close
+        </mwc-button>
+        <mwc-button slot="secondaryAction" dialogAction="close">
+          Just Close
+        </mwc-button>
+    </form>
+    </mwc-dialog>
     <script nomodule src="index.es5.js"></script>
   </main>
 </body>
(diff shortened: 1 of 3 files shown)


demos/dialog/index.html
```diff
@@ -155,6 +155,22 @@
         Secondary
       </mwc-button>
     </mwc-dialog>
+
+    <mwc-button data-num="8" raised>Validation in a Dialog</mwc-button>
+    <mwc-dialog id="dialog8" heading="Validation in a Dialog">
+      <div>
+        In this example we set validate a form inside a dialog before closing it.
```
e111077, line 162: please remove word "set"

packages/dialog/README.md
```diff
@@ -377,6 +377,45 @@ of the shadow root of `mwc-dialog` or the
 [root node](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode)
 of the dialog.
 
+### Form Handling
```
e111077, line 380: Can you make this more generic than "form handling". Something along the lines of conditionally closing dialog. This is especially since we're still working on how normal mwc elements will worl with forms.

e111077:
Spoke with the team and we want to be very specific about our wording here, so I'm going to close this issue as we have an engineer assigned to this. Though, we really do appreciate the effort and contribution

taylor-vann:
no worries! it's kinda why i was taking my time here. this seems like a very opinionated step.

YonatanKra:
I'll change the wording asap.
The example seems good to me as it is a valid use case. Or would you like
to avoid mentioning forms altogether?

PR #1560, opened 2020-07-22
ci(format:check): make format check output more useful

e111077:
see:

https://github.com/material-components/material-components-web-components/runs/900346141?check_suite_focus=true


e111077:
What do you mean producing diffs? `format:check` just runs `format` which in turn runs `clang-format` which is what we use both internally and externally. Problem is, I think internally has very slightly different flags than external (not to mention the text transformations we do)

allan-chen:
Diffs such as this, when running the command

```
@@ -199,6 +199,6 @@ export class CheckboxBase extends FormElement {
   }
 
   get isRippleActive() {
-    return this.rippleElement?.isActive || false;
+    return this.rippleElement ?.isActive || false;
   }
 }
```

e111077:
Hmm that doesn't look right. Are you running from a clean repo? is this copybara? if so, can you point me to the PR?

PR #1594, opened 2020-07-27
Remove crossorigin errors for preload links in Chrome.

taylor-vann:
Errors:

"The resource https://fonts.googleapis.com/css?family=Material+Icons&display=block was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally."

"The resource https://fonts.googleapis.com/css?family=Roboto:300,400,500 was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally."

Solution:
Can reproduce locally.
Removing the cross-origin attribute removes errors.
Tested:
- Chrome
- Firefox
- Safari


e111077:
please write up a bug for this for tracking

PR #1699, opened 2020-08-17
chore: remove src folder from packages in GitHub

asyncLiz:


asyncLiz:
Resolved offline, but they're purposefully not handled by copybara

PR #1844, opened 2020-09-25
mwc-checkbox: normalise tests

43081j:
This is just normalising the tests and adding a few assertions to match with what i did last time i reworked all the tests (added a "factory" kind of pattern).

### DIFF
--- a/packages/checkbox/test/mwc-checkbox.test.ts
+++ b/packages/checkbox/test/mwc-checkbox.test.ts
@@ -93,6 +93,11 @@ suite('mwc-checkbox', () => {
 
     test('does not animate after being hidden', async () => {
       element.checked = true;
+      const animatedElement =
+          element.shadowRoot!.querySelector('.mdc-checkbox__background')!;
+      await new Promise((resolve) => {
+        animatedElement.addEventListener('animationend', resolve);
+      });
       await element.updateComplete;
       element.style.display = 'hidden';
       await rafPromise();
--- a/packages/checkbox/test/mwc-checkbox.test.ts
+++ b/packages/checkbox/test/mwc-checkbox.test.ts
@@ -16,109 +16,163 @@
  */
 
 import {Checkbox} from '@material/mwc-checkbox';
+import {html} from 'lit-html';
 
-import {Fake, rafPromise} from '../../../test/src/util/helpers';
+import {Fake, fixture, rafPromise, TestFixture} from '../../../test/src/util/helpers';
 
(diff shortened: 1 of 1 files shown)


43081j:
FYI i ran tests on master too and got the same test failure so not too sure whats up.

e111077:
Heya, James, this looks generally good; there seems to be a CI failure in one test but I'm rerunning the job. Next I'm going to run it through internal coverage tests to give it a final check.

e111077:
If this tests fails again, try rebasing onto master?

e111077:
Seems like the same failure. Can you please give it a quick
`git checkout master && git pull && git checkout checkbox-tests && git rebase master`

43081j:
have rebased it now 👍  ill see if CI fails and try debug it a bit if it does. weirdly i was getting the same failure on my master branch so im hoping it just fixes itself tbh 🙈

43081j:
good news! it now passes on `master` locally but not on my branch. so ill try debug it a bit 

very weird since that test hasn't changed, the only change is that it uses a fixture now rather than `createElement` 🤔 

i wonder if before this test passed because of timings... we are now waiting longer maybe so `update` has run and set the animation class. im not entirely sure

e111077:
sorry this slipped through; I'll be able to take a closer look soon

43081j:
No rush I'll see if I can get chance to debug it too 👍 thanks for taking time to look

e111077:
Dan wrote this test and I think it just may have bugs. I think it's trying to do the following:

```diff
test('does not animate after being hidden', async () => {
      element.checked = true;
+     const animatedElement = element.shadowRoot!.querySelector('.mdc-checkbox__background')!;
+     await new Promise((resolve) => {
+       animatedElement.addEventListener('animationend', resolve);
+     });
+     await element.updateComplete;
      element.style.display = 'hidden';
      await rafPromise();
      element.style.display = '';
      await rafPromise();
      assert.equal(internals.animationClass, '');
    });
```

Write that into the test (but formatted) and that should fix the issue. I'll bring it up with him when Dan is back from vacation next week.

43081j:
i made the change and the tests pass for me now... but i have no idea still what that test is trying to test 😂

e111077:
I vaguely remember internally we were having an issue where on page load or when going from hidden to not hidden, the initial "check" animation would play every time. I think this was what it was testing in particular because it was added with the `@transitionend` callback

e111077:
FF failure. Rerunning to se if flake

Issue #1121, opened 2020-04-14
Is this project safe to adopt?


robertofabrizi:
I don't want to sound disrespectful or anything, I see the commits, but the "card" being a staple of MD as it is, and not being completed in a year, raises some doubts. 
The demo page referenced in the homepage is broken, also not a good sign...
What about components that are not even on the list, at least that I could see, like a date and time picker?
Thank you and be safe in these dark times.
Robert

abraham:
From the top of the readme:

> IMPORTANT: The Material Web Components are a work in progress and subject to major changes until 1.0 release.

rwestlund:
For what it's worth, I'm using this code in production for my company's SaaS app. It's not perfect, and I've had to implement a lot of stuff myself. Also a bunch of hacks, and reading the source of these elements to find out what they actually do. But I really needed to get off the old polymer elements (which I've almost done!), and there's not much else out there. I've been slowly transitioning as the elements here reach a point of being less buggy than the old polymer ones.

Google lost interest after the web components spec was adopted, and left this repository abandoned for a year, which is a shame. I'm grateful to the devs who have picked this back up, and I love how lightweight these components are. I hope they reach a point of maturity soon, because that will make my life and my code base much better, but this is clearly not a top priority for them.

If you're starting a new project, I would recommend going with react just for the more mature ecosystem. If you're already in the polymer or lit-html ecosystem, then I do recommend adopting this now. Just do it gradually, and expect problems.

I think lit-html is technically superior, and has a brighter future than react if it gets some more serious backing and adoption. Again, much gratitude to the devs here for the progress they've made.

Just my 2 cents.

e111077:
Hello, reiterating what abraham said, these elements are currently in alpha. More info can be found in my reply to #1080.

e111077:
@rwestlund I'm sorry you feel this way. When we started the Polymer Elements we were new to web components just as any other developer since it was unexplored territory. Unfortunately the team made a bunch of mistakes with our first set and also had run into issues with maintaining such a large set of components with such a small team. When the web components v1 spec was released we continued to iterate on web components usability via spec.

[...]

Issue #223, opened 2019-02-10
material-components-web vs material-components-web-components


gertcuykens:
Please correct me if I am wrong, doesn't the way you use web components break things like native capabilities with other elements for example the relation with `form` and `input` if you don't use the `is` attribute?

Aren't you pushing it a bit to far? It's very confusing to have two repo's named material-components-web and material-components-web-components. This feels like a Python 2.x vs 3.x struggle.

I understand material-components-web doesn't use the full potential of web components, but is it really worth the mess to duplicate the effort?

Strongly recommend to use the `is` attribute to ensure native capabilities or if browsers don't support it, drop it and join forces improving material-components-web

tionkje:
As far as i can tell material-components-web-components is a wrapper around material-components-web using web components. Not a reimplementation of the same thing.

gertcuykens:
Yes and no, for me promoting it like this would not be a reimplementation, but as far as I can tell they go far beyond this.

```html
<!doctype html>
<html>
  <head>
    <title>Randonneurs</title>
    <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
    <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes"/>
    <meta name="mobile-web-app-capable" content="yes">
    <style>
      html, body {
        padding: 0;
        margin: 0;
      }
    </style>
    <script src="https://unpkg.com/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
    <script type="module">
      import {html, render} from 'https://unpkg.com/lit-html'
      import {MDCTextField} from 'https://unpkg.com/@material/textfield?module'

      const helloWorld = () => {
        class HelloWorld extends HTMLElement {
          constructor() {
            super()
          }
          connectedCallback() {
            const root = this.attachShadow({mode: 'open'});
            render(html`
              <style>
                @import url('https://unpkg.com/material-components-web/dist/material-components-web.min.css')
              </style>
              <form>
                <div class="mdc-text-field">
                  <input type="text" id="my-text-field" class="mdc-text-field__input">
                  <label class="mdc-floating-label" for="my-text-field">Hint text</label>
                  <div class="mdc-line-ripple"></div>
                </div>
              </form>
            `, root)
            MDCTextField.attachTo(root.querySelector('.mdc-text-field'))
          }
        }
        customElements.define('hello-world', HelloWorld)
      }
      helloWorld()
    </script>
  </head>
[...]

PR #1852, opened 2020-09-29
Added accessibility features to mwc-textfield

christophe-g:
As a follow-up of #1215. 

The purpose of this PR is to make mwc-textfield more accessible, basically following https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield/helper-text#accessibility.

- https://github.com/material-components/material-components-web-components/commit/b020c1fa8dd29d0c21e8758b022b65fe7fcc0a20 makes test suite work on my machine. Test engine would not start without.

- https://github.com/material-components/material-components-web-components/commit/c3c45c17237c54abcefd7a505c7ab5b62746a932 is added to make use of chai a11y aXe testing lib (https://open-wc.org/testing/testing-chai-a11y-axe.html). The way it is embedded now is pretty ugly. Open to any better suggestions on how to hook this to the current testing library (I am not a all  used to typescript, so bare with me)

- main work done under https://github.com/material-components/material-components-web-components/commit/0693e313ccece935d524efe7db87974cbcf476fe. Tell me what needs to change before I commit similar work to textarea.


christophe-g:
Test failing on https://github.com/material-components/material-components-web-components/pull/1852/commits/0693e313ccece935d524efe7db87974cbcf476fe certainly because I used `pnpm` and not `npm` locally. I wish not to change this.

asyncLiz:
Thanks for the PR, and sorry for the delay on commenting!

We discussed this at our triage meeting last week, and our consensus is to hold off on implementing #1215 right now for a few reasons that are not immediately obvious:

1. Text field a11y must meet strict requirements internally and sync up to its internal implementation. While it may be possible to go back and forth on this PR right now to get it synced up, there are several outstanding bugs for the internal text field's a11y (some of which are being closed in a fixit this week), so it would quickly become out of sync. We'd like to find a better approach for keeping them in sync first or wait until the a11y story is more stable.
1. Design has stated that non-decorative actionable icons should display a ripple, since they are their own touch targets. Rather than adding more complex logic, we wish to implement this in MWC (and upstream in MDC) by using icon buttons. This would eliminate many of the requirements in the text field icon foundation and #1215 around icon a11y since that is built in to icon buttons already.
1. Changes to our build and test system are very difficult to bring in internally, and would require much more refactoring on this PR and effort that we don't think is beneficial right now for the short term given our current priorities.
1. The text field container element (which holds the icons) is expecting a refactor this quarter, which would make much of the above high effort work lost or redone.

We really appreciate the contribution, but given these reasons and our need to carefully upcoming work, we'd like to wait before addressing this outstanding bug.

christophe-g:
Thanks a lot for the detailed answer, truly appreciated !

Beyond the specifics of this PR, I think it illustrates a series of concerns I was not sure where/how to state. So, I'll do it here, not without thanking the devs who are putting tremendous amount of quality in the code base. 

1. It is difficult for me (us ?) to know where the library is heading and what is happening behind the scene. I am de-facto left with  proxy indicators at hands (like pace of development, type of PR being merged, milestones being pushed back) to guess what is happening. For the time being, they leave me intrigued. 

Would it make sense to post some type of high-level outline of medium term vision / problems encountered / area of work / what to expect in next releases ... It would be interesting to know a bit more about thinking behind theming options, envisioned long-term relationship with MDC foundation... It would also be kind of reassuring. For instance, discussion in #1121 was helpful. 

2. I understand the rationale behind "This repo is for extending the base layer with Web Components, and is not intended to be a complete replacement of MDC" (#223). I hope this is not on the expenses of loosing lit-element clarity and expressiveness and investing a large portion of the efforts just keeping in sync with upstream.

3. This is partially a consequence of 2 and potentially a costly one: it's hard to contribute! Except for very minor changes, it is often not clear where to intervene as the dependencies/multi-plexing with the base layer forces to digest multiple layers of overhead. 

I came to select this library because of the pleasant experience playing with / extending / contributing to other lit-based components (also because of very good experience with using Polymer elements). It wish it could be more of the same here. 

As a consequence I end up, like in the old times, with a series of fragile override for doing things the library does not yet care of.

I will now be closing this PR, as - again - reasons for not incorporating it makes sense, and open a couple follow-up issues (#1873, #1874).

packages/textfield/mwc-textfield-base.ts
```diff
@@ -387,6 +392,10 @@ export abstract class TextFieldBase extends FormElement {
     return html`
       <input
           aria-labelledby="label"
+          aria-controls="helper"
+          aria-describedby="helper"
+          aria-errormessage="helper"
```
christophe-g, line 397: @allan-chen - `aria-errormessage` and `aria-invalid`, ref for #1872

allan-chen: Thanks - I believe someone else is working on `aria-invalid` for textfield so I'll leave it out of scope of my PR for now. 

CC/ @brandondiamond - for your consideration we should add incorporate MWC into the `aria-invalid` enhancements, should be little overhead

PR #1863, opened 2020-10-01
upgrade typescript to 4.x

43081j:
is this useful? I was doing this while debugging a problem i had with duplicate tslib packages.

it doesn't look like it causes any changes in our own code, but would it still be considered a breaking change in mwc?


packages/base/base-element.ts
```diff
@@ -53,7 +56,9 @@ export abstract class BaseElement extends LitElement {
    *      ...
    *    }
    */
-  protected abstract createAdapter(): {}
+  protected abstract createAdapter(): {};
+
+  /* eslint-enable @typescript-eslint/ban-types */
```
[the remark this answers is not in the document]
e111077: I say go for it and check if build is fine. Though you may want to do this AFTER I check if this PR is feasible

e111077:
We may be blocked by google using TS 3.9.5. I will attempt to see if there are any immediate issues and whether this PR is feasible.

packages/menu/test/mwc-menu.test.ts
```diff
@@ -299,7 +299,7 @@ suite('mwc-menu', () => {
 
     setup(async () => {
       originalSetTimeout = window.setTimeout;
-      (window as any).setTimeout = (fn: Function) => {
+      (window as any).setTimeout = (fn: () => any) => {
```
e111077, line 302: does `unknown` work here. google style guide prefers over any

packages/snackbar/test/mwc-snackbar.test.ts
```diff
@@ -57,7 +57,7 @@ suite('mwc-snackbar', () => {
 
   setup(() => {
     originalSetTimeout = window.setTimeout;
-    (window as any).setTimeout = (fn: Function) => {
+    (window as any).setTimeout = (fn: () => any) => {
```
e111077, line 60: same as previous

e111077 (changes requested):

packages/menu/test/mwc-menu.test.ts
43081j: it'll work but we have to keep the `window as any` cast, i'll update

43081j:
do you want me to save my changes resolving your comments until you've tried it out through google as-is?

e111077:
it hasn't crashed yet and looks promising; do you mind resolving the comments? I was intending to do a release wednesday. If it makes it in after the release and it's not Friday, I'll try another release.

Also i think it would be good for this breaking change to be its own release anyway

43081j:
i pushed the resolved comments but didn't realise it also has the `createAdapter` change in it :facepalm:

if i need to revert that bit again just let me know

packages/menu/test/mwc-menu.test.ts
```diff
@@ -299,7 +299,7 @@ suite('mwc-menu', () => {
 
     setup(async () => {
       originalSetTimeout = window.setTimeout;
-      (window as any).setTimeout = (fn: Function) => {
+      (window as any).setTimeout = (fn: () => unknown) => {
```
e111077, line 302: another suggestion:

```ts
const timeoutShim: typeof window.setTimeout = (fn) => {
  (fn as ()=>unknown)();
  ...
  return -1
};
  
window.timeoutShim = a;
```

e111077 (changes requested):
I might end up fitting in a lit-element version bump to include `@query` caching before we submit this CL

packages/menu/test/mwc-menu.test.ts
43081j: yup i would prefer that too, much nicer to match the type. ill make the change

43081j: in actual fact this is proving to be difficult to do... as we have `@types/node` which is conflicting. somehow `setTimeout` is node's rather than the browser's.

is there no way we can not have the node types?

packages/snackbar/test/mwc-snackbar.test.ts
```diff
@@ -57,8 +57,10 @@ suite('mwc-snackbar', () => {
 
   setup(() => {
     originalSetTimeout = window.setTimeout;
-    (window as any).setTimeout = (fn: Function) => {
-      fn();
+    window.setTimeout = (fn) => {
+      if (typeof fn === 'function') {
+        fn();
+      }
```
43081j, line 63: this is what i ended up with @e111077. i hate the  condition here but we're in a predicament...

basically the linter is working against us in this case because `setTimeout` is typed as `fn: string|Function`. so if we annotate our types here or forcefully cast as `Function` we will hit the ESLint rule yet again (telling us off for using `Function` instead of a specific signature).

this seems to work fine but if we want to avoid having this condition we will have to ignore the lint rule. let me know what you think

e111077: why not just cast fn in the setTimeout as a function in TS? e.g. `(fn as ()=>unknown)();`?

43081j:
done all the comments now @e111077 but build seems to fail for some other error i haven't noticed before

e111077:
don't worry about the failure: temporary breakage pending on a sync with MDC canary

e111077:
fix submitted; you may have to rebase for these changes to propagate

43081j:
have rebased onto master if thats any help

PR #1928, opened 2020-11-02
Debounce the layout function in the list component

### LINKED ISSUE #1927: When list items are removed list.layout is called multiple times can causes a performance issue
**Describe the bug**
The list-item's disconnected callback triggers the list's layout which triggers the updateItems.  This causes a major performance issue with big lists.

**To Reproduce**
1. Go to https://codesandbox.io/s/cool-paper-0k530?file=/src/index.js
2. Open the developer tools in the performance tab and profile
3. Click on the `clear everything` button
4. Stop profiling
5. See that:
5.1 layout shift is reported
5.2 The task takes a huge amount of time (and blocks the UI for that amount of time)

**Expected behavior**
I'd expect the list items to be removed without a performance overhead.

**Screenshots**
That's how it looks in the "lab test" from the sandbox example:
image
This is how it looks in a real life example from an angular application using the mwc-select with mwc-list-item:
image

 - OS: MacOS
 - Browser Chrome, Brave

**Additional context
There's a need for this amount of items in the select. For instance - to show all the available timezones (around 400 timezones on earth apparently).

YonatanKra:
when removing an item.

Closes #1927 

See my remarks blow regarding the outcome of this performance improvement, and proposed changes to this PR.

### DIFF
--- a/packages/list/mwc-list-base.ts
+++ b/packages/list/mwc-list-base.ts
@@ -30,11 +30,9 @@ import {Layoutable, ListItemBase, RequestSelectedDetail} from './mwc-list-item-b
 export {ActionDetail, createSetFromIndex, isEventMulti, isIndexSet, MWCListIndex, SelectedDetail} from './mwc-list-foundation';
 
 function debounce(
-    callback: <T>(this: T, ...args: any[]) => void,
-    waitInMS = 50
-) {
+    callback: <T>(this: T, ...args: any[]) => void, waitInMS = 50) {
   let timeoutId: NodeJS.Timeout;
-  return function <T>(this: T, ...args: any[]) {
+  return function<T>(this: T, ...args: any[]) {
     clearTimeout(timeoutId);
     const context = this;
     timeoutId = setTimeout(() => callback.apply(context, args), waitInMS);
@@ -483,7 +481,6 @@ export abstract class ListBase extends BaseElement implements Layoutable {
   }
 
   debouncedLayout(): void {
-
   }
 
   layout(updateItems = true) {
--- a/packages/list/mwc-list-base.ts
+++ b/packages/list/mwc-list-base.ts
@@ -36,7 +36,8 @@ function debounceLayout(
     clearTimeout(timeoutId);
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     const context = this;
-    timeoutId = setTimeout(() => callback.apply(context, [updateItems]), waitInMS);
+    timeoutId =
+        setTimeout(() => callback.apply(context, [updateItems]), waitInMS);
   };
 }
 
@@ -44,9 +45,9 @@ const isListItem = (element: Element): element is ListItemBase => {
   return element.hasAttribute('mwc-list-item');
 };
 
-function clearAndCreateItemsReadyPromise() {
+function clearAndCreateItemsReadyPromise(this: ListBase) {
   const oldResolver = this.itemsReadyResolver;
-  this.itemsReady = new Promise(res => {
+  this.itemsReady = new Promise((res) => {
     return this.itemsReadyResolver = res;
   });
   oldResolver();
@@ -131,7 +132,10 @@ export abstract class ListBase extends BaseElement implements Layoutable {
   noninteractive = false;
 
   debouncedLayout: (updateItems?: boolean) => void | undefined;
-  itemsReadyResolver: (value?: (PromiseLike<never[]> | never[] | undefined)) => void = (() => {}) as (value?: (PromiseLike<any[]> | any[])) => void;
+  itemsReadyResolver: (value?: (PromiseLike<never[]>|never[]|undefined)) =>
+      void = (() => {
+                  //
+              }) as(value?: (PromiseLike<any[]>|any[])) => void;
 
   constructor() {
     super();
@@ -140,7 +144,7 @@ export abstract class ListBase extends BaseElement implements Layoutable {
       clearAndCreateItemsReadyPromise.call(this);
 
       debouncedFunction(updateItems);
-    }
+    };
   }
 
   itemsReady = Promise.resolve([]);
--- a/packages/list/mwc-list-base.ts
+++ b/packages/list/mwc-list-base.ts
@@ -34,6 +34,7 @@ function debounce(
   let timeoutId: NodeJS.Timeout;
   return function<T>(this: T, ...args: any[]) {
     clearTimeout(timeoutId);
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
     const context = this;
     timeoutId = setTimeout(() => callback.apply(context, args), waitInMS);
   };
@@ -67,7 +68,6 @@ export abstract class ListBase extends BaseElement implements Layoutable {
     }
   })
   activatable = false;
-
 
   @property({type: Boolean})
   @observer(function(this: ListBase, newValue: boolean, oldValue: boolean) {
@@ -121,6 +121,8 @@ export abstract class ListBase extends BaseElement implements Layoutable {
     }
   })
   noninteractive = false;
+
+  debouncedLayout: (updateItems?: boolean) => void | undefined;
 
   constructor() {
     super();
@@ -478,9 +480,6 @@ export abstract class ListBase extends BaseElement implements Layoutable {
     const target = e.target as ListItemBase;
 
     this.layout(this.items.indexOf(target) === -1);
-  }
-
-  debouncedLayout(): void {
   }
 
   layout(updateItems = true) {
--- a/packages/list/mwc-list-base.ts
+++ b/packages/list/mwc-list-base.ts
@@ -29,6 +29,17 @@ import {Layoutable, ListItemBase, RequestSelectedDetail} from './mwc-list-item-b
 
 export {ActionDetail, createSetFromIndex, isEventMulti, isIndexSet, MWCListIndex, SelectedDetail} from './mwc-list-foundation';
 
+function debounce(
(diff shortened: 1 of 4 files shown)


YonatanKra:
### Performance gains
The following measurements were taken for the test (test:debug, http://localhost:9876/debug.html?grep=mwc-list%3A%20mwc-list%20performance%20issue) before and after applying the change.
Before:
image
After:
image
Can see from the image that the remove process took a roughly 1/9 of the time after the change.

YonatanKra:
Caveat: The debounced updates are no longer part of the lists'updateComplete hook.

Can try to add it to the _getUpdateComplete somehow if needs be.

Another solution can be to allow the user to add a flag asyncRemoval or syncRemoval, and commence the layout accordingly.

Let me know if and how you'd like to proceed with this one.

packages/list/test/mwc-list.test.ts
```diff
@@ -1444,6 +1444,8 @@ suite('mwc-list:', () => {
         await items[1].updateComplete;
         await element.updateComplete;
 
+        await new Promise((res) => setTimeout(res, 50));
```
e111077, line 1447: This is going to cause lots of issues migrating users. Is there a way we can deterministically know when items is ready? Perhaps expose an `itemsReady: Promise<ListItem[]>` that gets chained by the debounce? (super skeleton code) e.g.

```ts
debounce(...) => {
  element.itemsReady.then(() => {fn(...);return element.items});
}
class {
  ...
  itemsReady: Promise.resolve([]);
}
```

packages/list/mwc-list-base.ts
```diff
@@ -29,6 +29,16 @@ import {Layoutable, ListItemBase, RequestSelectedDetail} from './mwc-list-item-b
 
 export {ActionDetail, createSetFromIndex, isEventMulti, isIndexSet, MWCListIndex, SelectedDetail} from './mwc-list-foundation';
 
+function debounce(
+    callback: <T>(this: T, ...args: any[]) => void, waitInMS = 50) {
```
e111077, line 33: can we please try to avoid `any` (`unknown` is okay for google style guide) since I doubt the fn defined in this file is going to be used somewhere else, we can be more specific with its implementation.

e111077 (changes requested):
please update `CHANGELOG.md`

e111077:
super thanks for the PR!

packages/list/test/mwc-list.test.ts
e111077: oh also you can override

```ts
async _getUpdateComplete() {
  await super._getUpdateComplete();
  await this.itemsReady;
}
```

so `list.updateComplete` should include `itemsReady`

e111077:
This seems like a simple enough implementation so let's go with this for now. Part of our code submission process includes importing this code into google's monorepo, and google-wide tests should help us run into migration issues other users may also have; this is where we can determine whether we need a flag. Though hooking into getUpdateComplete might be a requirement

YonatanKra:
@e111077 
Thanks for the comments. I've addressed all of them.
Let me know if something's missing.

packages/list/mwc-list-base.ts
```diff
@@ -113,6 +131,29 @@ export abstract class ListBase extends BaseElement implements Layoutable {
   })
   noninteractive = false;
 
+  debouncedLayout: (updateItems?: boolean) => void | undefined;
```
e111077, line 134: `protected debouncedLayout`

packages/list/mwc-list-base.ts
```diff
@@ -113,6 +131,29 @@ export abstract class ListBase extends BaseElement implements Layoutable {
...
+  itemsReadyResolver: (value?: (PromiseLike<never[]>|never[]|undefined)) =>
```
e111077, line 135: `protected itemsReadyResolver`

packages/list/test/mwc-list.test.ts
e111077, line 1447: Does this pass now if you remove this? if not, does it pass if you await itemsReady?

e111077 (changes requested):
final stretch! Great work so far!

packages/list/mwc-list-base.ts
YonatanKra: this cannot be protected as it is used outside the class (in the items `disconnectedCallback`

packages/list/test/mwc-list.test.ts
YonatanKra: Yes it does. Forgot to remove it - now it is removed.

e111077 (approved):

packages/list/mwc-list-base.ts
```diff
@@ -29,11 +29,30 @@ import {Layoutable, ListItemBase, RequestSelectedDetail} from './mwc-list-item-b
 
 export {ActionDetail, createSetFromIndex, isEventMulti, isIndexSet, MWCListIndex, SelectedDetail} from './mwc-list-foundation';
 
+function debounceLayout(
+    callback: <T>(this: T, updateItems: boolean) => void, waitInMS = 50) {
+  let timeoutId: NodeJS.Timeout;
```
e111077, line 34: This type is causing crashes internally. Could it be possible to cast it as `number` and you might also have to cast `clearTimeout` as `(id: number) => void`?

e111077 (changes requested):

packages/list/test/mwc-list.test.ts
```diff
@@ -1788,5 +1788,29 @@ suite('mwc-list:', () => {
...
+            const itemsTemplates = new Array(300).fill(0).map(() => listItem());
+            fixt = await fixture(listTemplate({items: itemsTemplates}));
+            element = fixt.root.querySelector('mwc-list')!;
+
+            count = 0;
+
+            fixt.remove();
+            await element.updateComplete;
+            fixt = null;
+            expect(count).to.eq(1);
```
e111077, line 1811: really dumb issue: please use `assert.equal` because internally we use Jasmine while out here we use karma and chai

e111077 (changes requested):

packages/list/mwc-list-base.ts
```diff
@@ -29,11 +29,31 @@ import {Layoutable, ListItemBase, RequestSelectedDetail} from './mwc-list-item-b
 
 export {ActionDetail, createSetFromIndex, isEventMulti, isIndexSet, MWCListIndex, SelectedDetail} from './mwc-list-foundation';
 
+function debounceLayout(
+    callback: <T>(this: T, updateItems: boolean) => void, waitInMS = 50) {
```
e111077, line 33: YouTube conformance issue basically boils down to please use a different argument name

packages/list/mwc-list-base.ts
```diff
@@ -29,11 +29,31 @@ import {Layoutable, ListItemBase, RequestSelectedDetail} from './mwc-list-item-b
...
+  let timeoutId: number;
+  return function<T>(this: T, updateItems = true) {
```
e111077, line 35: same as above

e111077 (changes requested):

packages/list/mwc-list-base.ts
YonatanKra: Simplified it. Hope it's ok now :)

packages/list/mwc-list-base.ts
YonatanKra: What argument? updateItems? callback?

e111077: hahaha oops I meant `this`

YonatanKra: oh - `this`?
I've removed it - all tests pass, linting seems good :)

YonatanKra:
Resolved all mentioned issues 😀

e111077 (approved):

packages/list/test/mwc-list.test.ts
```diff
@@ -1788,5 +1788,32 @@ suite('mwc-list:', () => {
...
+      test(
+          'removing a list should not call layout more than once', async () => {
+            let count = 0;
+            const originalLayout = List.prototype.layout;
+            List.prototype.layout = function(update) {
+              originalLayout.call(this, update);
+              count++;
+            };
+
+            const itemsTemplates = new Array(300).fill(0).map(() => listItem());
```
e111077, line 1802: Not sure why our IE tests are not running externally, but can we decrease this to 150? Does that still test correctly? IE is timing out and taking too long (>5s) to process 300 components.

e111077 (changes requested):
I'm also debugging a failure in a YouTube test that uses select. A little afraid it may be caused by the debounce vs a buffer, but it's still too early to tell

YonatanKra:
> I'm also debugging a failure in a YouTube test that uses select. A little afraid it may be caused by the debounce vs a buffer, but it's still too early to tell

It is a breaking change, as it changes the interface (from sync to async). So it might be this test assumes the remove event should have sync effect.

packages/list/test/mwc-list.test.ts
YonatanKra: I believe the issue is related lit-element. It also has a performance issue when adding a lot of elements to the DOM (even in chrome). 
In the profile I've added to this PR you can see the `_enqueueUpdate` that's taking a lot of time to load the fixture.

YonatanKra: Digging a bit deeper, it seems like lit-html and lit-element are being called a lot of times by the`list` element.
That's a different issue regarding the addition of elements to the list.

e111077:
Haven't forgotten about this; still trying to untangle their esoteric code

PR #1956, opened 2020-11-13
Uprade to typescript 4x

taylor-vann:
Migration from 0.19.1 -> 0.20.0


e111077 (approved):

asyncLiz:
If this PR is updating typescript, I would expect to see a change for the root [package.json](https://github.com/material-components/material-components-web-components/blob/master/package.json#L97) bumping the typescript dependency

e111077:
IIRC this is a fix of an issue in #1863

dfreedm:
Let's not rev package version with the typescript update. Also, this needs to be sync'd against master

e111077:
Wait, something weird is going on...

#1863 was merged via copybara which has a change to `/package.json` to update ts to 4.0. All other changes (the ones in google3 🙄) seem to have merged but not the one to the `package.json` which is in the root.

This CL was just meant to initially prepare for release and update the changelog, but Liz noticed that `/package.json` did not have an updated `typescript` dependency.

What needs to happen is we need to update the TS dependency in the root package without Copybara and then use this (or something like this) PR to ready for release

taylor-vann:
Closing for sanity reasons.
Will make separate PR for package json Typescript bump.
Will make separate PR after for new release.

PR #2024, opened 2020-12-21
chore: update status dates

e111077:
We've had some confusion from our community on these dates and they are outdated


asyncLiz (approved):
lgtm

PR #2063, opened 2021-01-11
chore: bump safari test versions & fix test

### LINKED ISSUE #1869: Tests should run in latest Safari
It looks like tests only run in Safari 10 and 11? The latest version (14) should also be tested.

https://github.com/material-components/material-components-web-components/blob/master/karma.conf.js#L110

e111077:
Bumped safari versions, for some reason transitionend does not fire on safari 13 twice in a row. Behavior seems correct, but the browser is just not firing the event. Safari 13-specific workaround implemented

fixes #1869

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -104,24 +104,18 @@ module.exports = function(config) {
         version: '11',
         platform: 'Windows 8.1',
       },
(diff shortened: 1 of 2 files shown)


asyncLiz (approved):
LGTM

PR #2098, opened 2021-01-22
Fix: mwc-menu min-width according to material-ui spec

### LINKED ISSUE #2096: [menu] fix wrong min-width
**Describe the bug**

Min width on menu is auto but spec says it should be 112px. Need to migrate users over

**Screenshots**

https://material.io/components/menus#specs
image

Shinyaigeek:
close #2096 

## Problem

> Min width on menu is auto but spec says it should be 112px. Need to migrate users over

## Screenshots

|before|after|
|:--:|:--:|
|スクリーンショット 2021-01-22 22 41 34|スクリーンショット 2021-01-22 22 40 56|


Shinyaigeek:
@googlebot I signed it!

Shinyaigeek:
@googlebot I signed it!

packages/menu/mwc-menu.scss
```diff
@@ -21,5 +21,5 @@ mwc-list ::slotted([mwc-list-item]:not([twoline])) {
 
 mwc-list {
   max-width: var(--mdc-menu-max-width, auto);
-  min-width: var(--mdc-menu-min-width, auto);
+  min-width: var(--mdc-menu-min-width, 112px);
```
e111077, line 24: Hello thanks for the submission! Unfortunately, there has been a refactoring of this code and min and max-width have been moved to menu surface So there are a few things that would be useful here:

1. please make this change in `mwc-menu-surface.scss`
2. please use the `menu-mixins.min-width` mixin found [here](https://github.com/material-components/material-components-web-components/blob/master/packages/menu/mwc-menu-surface.scss#L33)
3. please use [`menu-variables.$min-width`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/_variables.scss#L32) instead of hard coding `112px` (it seems we forgot to clean up the import, oops, but that makes it easier for you)

Also we have lots of screenshot tests internally, so I'll have to do some work behind the scenes migrating google to get this PR in, so don't worry if you see this PR just sitting around for a bit – it's actively being worked on

e111077 (changes requested):

packages/menu/mwc-menu.scss
Shinyaigeek: Hello ! Thank you for your excellent detailed review !
I did not notice refactoring, I’m sorry :pray:
I fixed my code according to what you pointed out :pray:

> Also we have lots of screenshot tests internally, so I'll have to do some work behind the scenes migrating google to get this PR in, so don't worry if you see this PR just sitting around for a bit – it's actively being worked on

OK, I am looking forward to this PR merged :+1:

e111077 (approved):

packages/menu/mwc-menu.scss
e111077: no worries it was merged after your PR

PR #2147, opened 2021-02-06
Allows setting a custom menu corner property to the select element

vpalmisano:
Adds a `menuCorner` property to the select element for setting the internal menu corner property to a custom value, different from the current `BOTTOM_START`.

### DIFF
--- a/packages/select/mwc-select-base.ts
+++ b/packages/select/mwc-select-base.ts
@@ -30,6 +30,7 @@ import {floatingLabel, FloatingLabel} from '@material/mwc-floating-label';
 import {lineRipple, LineRipple} from '@material/mwc-line-ripple';
 import {ListItemBase} from '@material/mwc-list/mwc-list-item-base';
 import {Menu} from '@material/mwc-menu';
+import {Corner} from '@material/mwc-menu/mwc-menu-surface-base';
 import {NotchedOutline} from '@material/mwc-notched-outline';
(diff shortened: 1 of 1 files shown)


dfreedm:
What's the use case for changing the corner? Do you have a mock of what you would do with other corners?

vpalmisano:
For example, to obtain this result:

screenshot-10 0 0 137_8000-2021 02 13-12_19_10

asyncLiz:
Thanks for the mockup! Unfortunately we've discussed this option with our designers and a11y engineers, and they do not recommend covering the text field with the menu. Adding an option for the user to break this convention isn't something we'd like to support, but extending our components and adding the option yourself is always a valid workaround!

vpalmisano:
Ok, I understand but the same feature it is already implemented in Angular:
https://material.angular.io/components/select/overview#getting-and-setting-the-select-value

and React:
https://material-ui.com/components/selects/

asyncLiz:
Our goal on the Material Design team is to implement Material components to the exact specifications and highest accessibility standards required by our designers. While this results in high quality and accessible components that are the gold standard for Material spec compliance, it sometimes means less built-in flexibility to break those design and accessibility standards.

In these situations, our recommendation is to extend our components for products with requirements that we are not able to implement due to requirements of our own. We do our best to ensure our components are flexible enough to be reasonably extended for supporting features after the fact!

PR #2177, opened 2021-02-18
Standardize mwc-ripple import style

### LINKED ISSUE #2176: Can't import both mwc-button and mwc-tab from unpkg.com
**Describe the bug**

Importing both `@material/mwc-button` and `@material/mwc-tab` from unpkg.com causes `mwc-ripple` to be registered twice, throwing an exception.

I think this is because they use two different styles for importing `mwc-ripple.js`. Even though unpkg.com redirects these to the same URL, module identity in the browser is always based on the original pre-redirected URL, so they are considered two modules.

`mwc-button.js` has:

```js
import '@material/mwc-ripple/mwc-ripple';
```

While `mwc-tab.js` has:

```js
import '@material/mwc-ripple';
```

**To Reproduce**

https://jsbin.com/vukucisoki/edit?html,console

```html
<script type="module" src="https://unpkg.com/@material/mwc-button?module"></script>
<script type="module" src="https://unpkg.com/@material/mwc-tab-bar?module"></script>
```

```
"Error: Failed to execute 'define' on 'CustomElementRegistry': the name \"mwc-ripple\" has already been used with this registry
    at legacyCustomElement (https://unpkg.com/lit-element@2.4.0/lib/decorators.js?module:15:25)
    at https://unpkg.com/lit-element@2.4.0/lib/decorators.js?module:50:1
    at __decorate (https://unpkg.com/tslib@2.1.0/tslib.es6.js?module:58:88)
    at https://unpkg.com/@material/mwc-ripple@0.20.0/mwc-ripple.js?module:25:10"
```

**Expected behavior**
No error

aomarks:
This avoids problems with unpkg.com (and maybe elsewhere too) where the use of two different import styles for the same module causes the double registration of mwc-ripple. This is because module identity in the browser is based on the pre-redirected URL.

Fixes #2176

### DIFF
--- a/packages/button/mwc-button-base.ts
+++ b/packages/button/mwc-button-base.ts
@@ -15,9 +15,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import '@material/mwc-icon/mwc-icon';
-import '@material/mwc-ripple';
+import '@material/mwc-ripple/mwc-ripple';
 
-import {Ripple} from '@material/mwc-ripple';
+import {Ripple} from '@material/mwc-ripple/mwc-ripple';
 import {RippleHandlers} from '@material/mwc-ripple/ripple-handlers';
 import {eventOptions, html, internalProperty, LitElement, property, query, queryAsync, TemplateResult} from 'lit-element';
 import {classMap} from 'lit-html/directives/class-map';
--- a/packages/button/mwc-button-base.ts
+++ b/packages/button/mwc-button-base.ts
@@ -15,9 +15,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import '@material/mwc-icon/mwc-icon';
-import '@material/mwc-ripple/mwc-ripple';
+import '@material/mwc-ripple';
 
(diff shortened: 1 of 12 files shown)


packages/button/mwc-button-base.ts
```diff
@@ -15,9 +15,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import '@material/mwc-icon/mwc-icon';
-import '@material/mwc-ripple/mwc-ripple';
```
asyncLiz, line 18: Our style guide recommends importing directly from files instead of file directory bundles (avoids copybara directory name change issues with Sass), so the opposite should be the case here. If there are any instances of `import '@material/mwc-ripple'`, they should point to `import '@material/mwc-ripple/mwc-ripple'`

asyncLiz (changes requested):

packages/button/mwc-button-base.ts
aomarks: Done.

asyncLiz (approved):
LGTM

Issue #2133, opened 2021-02-02
Allow to set buttons border radius


maicol07:
**Is your feature request related to a problem? Please describe.**
At the moment standard buttons border radius is set to 4px and it's impossible to change via CSS.

**Describe the solution you'd like**
Allow to change the border radius of the mwc-button component. This will allow us to get rounded buttons, like this one:
![immagine](https://user-images.githubusercontent.com/9463142/106593743-7d11b180-6551-11eb-959b-e1e7efad4a81.png)

**Describe alternatives you've considered**
Using the ::shadow CSS selector but it doesn't seem to work...
The only workaround I found is using JS:
```js
document.querySelector('.mwc-button--rounded').shadowRoot.firstElementChild.style = 'border-radius: 100px;';
```

taylor-vann:
Hey @maicol07

Thanks for submitting. We have custom properties that affect 'shape' similar to your example :)

Unfortunately our docs need to be updated to include 'shape' in our theming :(

I'll fix the docs and update here!

asyncLiz:
Fixed in https://github.com/material-components/material-components-web-components/pull/2179

PR #2179, opened 2021-02-18
Theme docs with shape

taylor-vann:
Heyyo!

Added an example of shape to the theming docs!

#2133


docs/theming.md
```diff
@@ -217,18 +217,23 @@ html {
...
+  --mdc-shape-large: 16px;
+}
+```
+
 MWC doesn't support cut corners.
 
-See also: [Material Design shape](https://material.io/design/shape/applying-shape-to-ui.html#).
--->
+See also: [Material Design type system](https://material.io/design/shape/applying-shape-to-ui.html#shape-scheme)
+
```
asyncLiz, line 236: This should still be "shape" and missing a period. Also remove the extra newline so there's only one empty line between the sentence and the next section.

```suggestion
See also: [Material Design shape](https://material.io/design/shape/applying-shape-to-ui.html#shape-scheme).
```

asyncLiz (changes requested):

docs/theming.md
taylor-vann: gooood catch

taylor-vann (commented):
Resolved in the following commit :)

docs/theming.md
taylor-vann: resolved in the next commit

asyncLiz (approved):
LGTM

PR #2187, opened 2021-02-22
chore: update circular progress tests to use fixtures

43081j:
i remember we were moving tests towards this structure so figured i'd do a few more.

i've updated the circular progress bar and the four colour one, too. now we cover the defaults and also the behaviour of setting out of bounds progress percentages.

let me know if this is fine & still the way you're writing them

cc @e111077

### DIFF
--- a/packages/circular-progress-four-color/test/mwc-circular-progress-four-color.test.ts
+++ b/packages/circular-progress-four-color/test/mwc-circular-progress-four-color.test.ts
@@ -15,61 +15,117 @@
  * limitations under the License.
  */
 import {CircularProgressFourColor} from '@material/mwc-circular-progress-four-color';
+import {html} from 'lit-html';
+
+import {fixture, TestFixture} from '../../../test/src/util/helpers';
 
 const INDETERMINATE_CLASS = 'mdc-circular-progress--indeterminate';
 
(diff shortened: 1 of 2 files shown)


e111077 (approved):

e111077:
hmm IE tests fail the set aria-label test on the initialization check. Investigating

e111077:
seems to be, no matter how long I await, the attribute ariaLabel is not being turned into a prop. I went ahead and fixed it on my end by changing `ariaLabel=${propsInit.ariaLabel ?? ''}` to `.ariaLabel=${propsInit.ariaLabel ?? ''}`

This is an issue for further investigation later

43081j:
specifically in IE? I wonder if they reserve `arialabel` or something. could try setting it on a div in IE and see if it lets you (I don't have IE 😬 )

good idea on the fix anyhow, thanks for merging :)

e111077:
I fixed it internally which a bot merges in a separate CL with my change which then merges this CL because they share the same hash, so already fixed in master

43081j:
@e111077 FYI i did try it in IE and it does _not_ let you name an attribute `arialabel`. explains a lot

e111077:
Oof do you mind filing a bug

PR #2293, opened 2021-04-02
test(fab): rework tests to use fixtures

43081j:
have done the same i did with the other tests:

- moved them to use fixtures
- added assertions for the initial values
- added tests for `reducedTouchTarget`

cc @e111077

### DIFF
--- a/packages/fab/test/mwc-fab.test.ts
+++ b/packages/fab/test/mwc-fab.test.ts
@@ -29,6 +29,7 @@ interface FabProps {
   showIconAtEnd: boolean;
   icon: string;
   label: string;
+  reducedTouchTarget: boolean;
 }
 
 const defaultFab = html`<mwc-fab></mwc-fab>`;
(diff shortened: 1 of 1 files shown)


dfreedm (approved):
LGTM. Throwing onto the internal testing train

43081j:
Looks like tests failed in the internal pr. Can try take a look tomorrow unless you already have an idea of what's up

dfreedm:
Thankfully its failing on IE11, not something internal-only. Our Saucelabs runner only runs for PRs on branches in the repo.

43081j:
@dfreedm  i suspect this is because ie11 is unhappy with camel case attributes and mapping them to properties, somehow.

i will update to use properties for those, but the intention was to assert that attributes set properties correctly, so we would lose that.

43081j:
@dfreedm @e111077 could one of you re-run the internal CI stuff when you get chance so we can merge this? 🙏 

it should pass now

dfreedm:
Looks good. We have a slightly different way of handling dependencies internally, and because of some quirks with our github sync system and manual internal fixes, this will land in #2413 with no files changed 🤷

PR #2357, opened 2021-04-29
fix(mwc-menu): issue 2356 fixing menu body mis-click

### LINKED ISSUE #2356: [mwc-menu] menu is not closed when document clicked
**Describe the bug**
In the pure simply plain HTML the body is not being stretched over the whole document, it is rather stay small and only extends as much as layouted elements in it.
When `mwc-menu` is opened in such a setup, clicking on most of the area of the viewport won't close the menu.

**To Reproduce**
1. Create simple HTML file
2. Import `mwc-menu`, set it with few list items and put in the body
3. Open menu
4. Click somewhere in the lower half of the screen
5. Observe that menu is not closed
6. Click somewhere at the top of the view
7. The menu closes as expected

**Expected behavior**
Menu should be closed on click wherever user clicks

Suggestion:
Now you are registering the `onBodyClick` in the `mwc-menu-surface` component on the `document.body`.
https://github.com/material-components/material-components-web-components/blob/5dd85d0c6700033fdc0da40d2424e9e8c46b0141/packages/menu/mwc-menu-surface-base.ts#L346

Do register it on the document instead.

gullerya:
closes #2356

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "sass": "^1.24.4",
         "shady-css-parser": "^0.1.0",
         "tachometer": "^0.4.13",
-        "typescript": "^4.0.3"
+        "typescript": "^4.1.4"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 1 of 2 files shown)


gullerya:
@googlebot I signed it!

PR #2367, opened 2021-04-30
fix: compat between lit-element 2.4.0 & 2.5.0

e111077:
2.5.0 has a compatibility change to lit 1 for lit 2. Since we were mid-migration to Lit 2, we doubled up on these efforts and caused an infinite loop.

cc: @dfreedm @bicknellr


e111077:
Git diff gore

Issue #2348, opened 2021-04-27
Modernize Sass build process


dfreedm:
## Description
Our sass build process is very custom and awkward. We should have a per-package build that only needs to rebuild styling for modified components, instead of a global "compile everything" pass.

## Acceptance criteria
Styling should be able to be built on a per-component basis, and run in parallel.

dfreedm:
Fixed with #2421 and #2418

PR #2418, opened 2021-05-17
New sass compile pass

dfreedm:
- Reuse `sass-render` internals, but structured so that each component
  can build independently
- Emit stylesheet as `styles` rather than `style` to match internal
  build
- Emit stylesheets as `.css.ts` instead of `-css.ts` to match internal
  build

Related #2348

### DIFF
--- a/scripts/sass-to-lit-css/index.js
+++ b/scripts/sass-to-lit-css/index.js
@@ -1,0 +1,49 @@
+#!/usr/bin/env node
+
+/**
+@license
+Copyright 2021 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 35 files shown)


dfreedm:
I'll have a followup PR to switch `npm run build:styling` to use this new path with all the components soon

scripts/sass-to-lit-css/index.js
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env node
+
+/**
+@license
```
e111077, line 4: mind just opening a quick issue / task to migrate all of our license headers to the new, shorter format?

e111077 (approved):
pre-approval

scripts/sass-to-lit-css/index.js
dfreedm: Filed as #2419. I'll update this template to use the SPDX header, and follow up with the rest of the code in a separate PR.

PR #2421, opened 2021-05-18
(change): Import styles from `.css.ts` instead of `-css.ts`

dfreedm:
(change): Import styles from `.css.ts` instead of `-css.ts`


dfreedm:
Related to #2348

PR #2424, opened 2021-05-18
feat(tab): Add min-width css variable

bnf:
Allow to configure the default 90px min-width that has been added in
https://github.com/material-components/material-components-web/commit/c4ab98722

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- `tab`
+  - Add `--mdc-tab-min-width` variable
+
 ## [v0.21.0] - 2021-04-30
 
 ### Changed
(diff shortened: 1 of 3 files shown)


e111077:
Hello, thanks for the PR!

We are currently in the process of implementing Material Design Tokens across the MDC and MWC libraries (see fab for example). We are currently going from a top-down design team to us approach on this, so I don't know if min-width will be a supported token then. If it is not, then we would have to remove this feature in the future. If it is, then the SASS infrastructure under the hood will have to change for this component.

We currently do not want to provide a feature that may be removed in the following months. A workaround for now would likely be to subclass the tab component and add that in your styles. e.g.

```ts
class MyTab extends Tab {
  static styles = [...Tab.styles, css`insert custom prop here`].
}
```

Though you may want to wait until [this PR](https://github.com/material-components/material-components-web-components/pull/2427) merges as it will standardize our style subclassing strategy

PR #2448, opened 2021-05-25
chore: Remove typescript lint from GitHub tests

abhiomkar:
External typescript lint (including format lint) conflicts with internal TS lint and produces different results. This impedes devex because the turn around time to validate lint is slow (i.e., waits till GitHub Action is complete), also has no integration to internal review tool.

All the code that gets checked-in should be tested internally.


dfreedm:
I'm not 100% sure that the internal typescript linting is enabled for this repo, but it seems ok to me to turn off the eslint checks. There's quite a few dependencies we could drop that way.

dfreedm (approved):

PR #2478, opened 2021-06-14
Add `mwc-base` dependency to linear progress

bramkragten:
`mwc-base` is used in linear progress, but was missing from its dependencies.


e111077:
Thanks for the submission. Just made a quick merge conflict fix

PR #2541, opened 2021-07-15
build: Remove external format checks

abhiomkar:
Removed external format checks to stay in sync with internal format.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,9 +34,6 @@ jobs:
     - name: Lerna bootstrap
       run: npm run bootstrap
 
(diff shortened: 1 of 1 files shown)


asyncLiz (approved):
LGTM, but how come clang-format isn't in sync with internal TS formatting? Our config is just BasedOnStyle: Google. Do we need to update the dependency?

abhiomkar:
TS Auto format is now enforced internally. I'll hold off merge this PR.

dfreedm:
I do like the idea to at least unblock unit test runs from format checks. I'll do that in a separate PR

PR #2576, opened 2021-07-26
chore(slider): allow dependency updating and update demo

e111077:
does not need copybara. external only


taylor-vann (approved):
looks great! thanks elliot!

PR #2584, opened 2021-07-29
Skypack fixes

aomarks:
I tried importing every MWC element from skypack, and found a few issues.

- Fix broken `elevation-overlay` package `main` and description
- Fix missing `package.json` dependencies on `@material/mwc-base` and `lit-html`. Skypack requires that every import has an explicit dependency in the `package.json` -- transitive dependencies are not sufficient.
- Normalize import style for `mwc-ripple` and `mwc-icon`. This should help reduce duplicate registration errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - **BREAKING:** `slider`
   - Initial implementation of Material 2 slider
   - Implementation is vastly different from M1 slider
+- `list`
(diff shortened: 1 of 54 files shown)


e111077 (approved):

e111077:
The change in `@material/mwc-ripple/mwc-ripple` to `@material/mwc-ripple` breaks L2W. I don't think we'll be able to stop the duplications.

e111077:
If you update the PR to fix only the main and module fields I'll be able to merge that

e111077 (changes requested):
remove import fixes

PR #2598, opened 2021-08-05
chore: exclude deprecated switch from tests

asyncLiz:
Needed to merge https://github.com/material-components/material-components-web-components/pull/2591


asyncLiz:
This will temporarily break master and be fixed after mwc-switch update is merged in. Can't be done at the same time since this file isn't synced

e111077 (approved):

e111077:
Elizabeth wants an LGTM but doesn't want a merge rn

PR #2652, opened 2021-08-26
feat(mwc-select): add support for slotted icons

### LINKED ISSUE #1310: Select field should allow for custom icons
**Is your feature request related to a problem? Please describe.**
For users who don't want to rely on an icon font it would be great if the `mwc-select` component could support external icons via slotting — much like the `mwc-button` component.

**Describe the solution you'd like**
If a user adds an icon inside the `mwc-select` component by using the `slot="icon"` attribute, it should render the icon in the position where icons rendered using the `icon` attribute would be rendered.
```html
<mwc-select>
  <svg slot="icon">
    ...
  </svg>
   ...
</mwc-select>
```

**Describe alternatives you've considered**
Currently, there is no other way I know of to get non icon font icons in the `mwc-select`.

**Additional context**

afberg:
Closes #1310

### DIFF
--- a/packages/select/README.md
+++ b/packages/select/README.md
@@ -204,7 +204,7 @@ focused:
 Name      | Description
 --------- | ----------------------------------------------------------------
 _default_ | Content to display in the selects internal `<mwc-menu>` element.
-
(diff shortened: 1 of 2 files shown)


asyncLiz:
Obsolete with M3

PR #2661, opened 2021-08-30
Provide variable to user for changing dialog padding

itsmeneartou:
I recently faced an [issue](https://github.com/material-components/material-web/issues/2660) that I wasn't able to change the padding of dialog due to shadow-root. So I create variables to change padding they are named as follows `--mdc-dialog-padding-top`, `--mdc-dialog-padding-right`, `--mdc-dialog-padding-bottom`, `--mdc-dialog-padding-left`.

### DIFF
--- a/packages/dialog/mwc-dialog.scss
+++ b/packages/dialog/mwc-dialog.scss
@@ -93,13 +93,7 @@
   }
 
   .mdc-dialog__content {
-    @include mwc-dialog-padding(
-      dialog.$padding-top,
-      dialog.$padding-right,
-      dialog.$padding-bottom,
-      dialog.$padding-left,
-      $query: $query
-    );
+    @include mwc-dialog-padding(20px, 24px, 20px, 24px, $query: $query);
   }
 }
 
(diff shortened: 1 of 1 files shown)


itsmeneartou:
@googlebot I signed it!

e111077:
Unfortunately we are in the middle of a rewrite to support Material Tokens. This comes from a pipeline in which the tokens are not yet fully baked.

We may consider a PR that follows the structure and style of `switch` e.g.

[`_switch-theme.scss`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-switch/_switch-theme.scss) & [`_switch.scss`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-switch/_switch.scss) in the mdc repo and [`_switch-theme.scss`](https://github.com/material-components/material-web/blob/master/packages/switch/_switch-theme.scss) & [`_switch.scss`](https://github.com/material-components/material-web/blob/master/packages/switch/_switch.scss) in the MWC repo.

If you're willing to make this refactor let us know and I can give you the token names and reopen the PR. If not, we will make this change in due time.

PR #2680, opened 2021-09-08
fix(mwc-textfield): hide element if type is hidden

yinonov:
feel free to refactor or modify but I think the hidden style is neccessary to hide the element as a whole

### DIFF
--- a/packages/textfield/mwc-textfield.scss
+++ b/packages/textfield/mwc-textfield.scss
@@ -28,6 +28,15 @@
   outline: none;
 }
 
+:host([type="hidden" i]) {
+  display: none;
+  appearance: none;
+  background-color: initial;
+  cursor: default;
+  padding: initial;
+  border: initial;
+}
+
 .mdc-text-field {
   width: 100%;
 


asyncLiz:
What is your use case for using `<mwc-textfield type="hidden">` instead of `<input type="hidden">`?

yinonov:
I agree that using the input will have the same effect and a very simple workaround, but what if the textfield is dynamically displayed?
you can obviously have plenty of workaround to that use-case but wouldn't it be aligned with input just for the little cost of it?

asyncLiz:
I think we'd rather avoid adding maintenance for use cases that can be solved in alternate ways.

For dynamically displayed text fields, I would imagine that the actual use case for a text field that is hidden and then displayed would be controlled with CSS (either `opacity: 0` with a transition or `display: none`/`transform`). Alternatively it can be conditionally rendered within the framework such as with lit-html.

`<input type="hidden">` is commonly used for form submission to track values that should not be visible to the user. I don't believe I've seen many common use cases of toggling the input's type to control its visibility, since changing the type can change other things such as its validity.

yinonov:
hmm fair enough. closing

Issue #2742, opened 2021-09-22
v0.25.1 components fail to execute directives - and breaks template - when used with other lit-2 based components


christophe-g:
**Describe the bug**

`mwc-switch` fail to execute directives (classMap and ifDefined) because it loads them from pre-lit-2 locations. 

Rendering those elements (I am running dev mode with vite), the DOM tree looks like: 

![Selection_212](https://user-images.githubusercontent.com/1168053/134403309-96dcac61-8a91-411d-8ce0-92b784962b14.png)

The root cause of this issue is that mwc-switch loads directives from `lit-html` and register duplicate version of the directives. 

Those lines in `mwc-switch-base` 
```js
import {ClassInfo, classMap} from 'lit-html/directives/class-map';
import {ifDefined} from 'lit-html/directives/if-defined';
```

Should be (tested as fixing this issue)
```
import {ClassInfo, classMap} from 'lit/directives/class-map.js';
import {ifDefined} from 'lit/directives/if-defined.js';
```

This problem is related and similar to #2736, but more severe as it breaks the app.

christophe-g:
It is actually affecting almost all components. 

I guess this behavior is not observed in https://material-components.github.io/material-web/demos/index.html or in tests because mwc-components are all loading their directive from the same source (lit-html/directives). 

The situation is different as soon as you use them along with other lit-based components updated as per Lit upgrade guide: https://lit.dev/docs/releases/upgrade/#update-packages-and-import-paths.

asyncLiz:
This should be fixed in #2749

christophe-g:
Thanks for this - will there be a patch release including this fix  (or is @canary the way to go)?

PR #2747, opened 2021-09-23
fix import of lit directives from lit-html/directives to lit/directives

christophe-g:
A fix for #2742


asyncLiz:
Thanks for the PR! We're actually working on a fix for this now in #2749, which requires some internal updates as well. Feel free to track that for an update :)

Issue #2759, opened 2021-09-25
[mwc-textfield] breaks UI if `type` is set to 'submit', 'reset', 'button'


yinonov:
**Describe the bug**
the types values of 'submit', 'reset', 'button' will render the native input element as a button, therefore breaks the UI and place a button instead of the expected `contenteditable` input.

**To Reproduce**

**Expected behavior**
The above mentioned types should be restricted / ignored or fallback to a default value and keep textfield as a `contenteditable` input.

**Screenshots**
![image](https://user-images.githubusercontent.com/15140652/134765970-a8a8db00-0de8-4971-b1d0-53038b13ea31.png)

dfreedm:
Thanks for the bug report! 
Given that this looks _obviously_ broken, and that there are probably very few use cases for `submit`, `reset` or `button` types, I think we can just add some documentation and Typescript type validation.

yinonov:
Fair enough!

schirrel:
@dfreedm do you think would be a good approach to throw an error when using these types? Once they are not int `TextFieldType` and are being used at `type` property:
``` @property({type: String}) type: TextFieldType = 'text';```
Or maybe at the `renderInput` return the mwc-button?

asyncLiz:
M3 now has explicit supported and unsupported text field types that are enforced with typescript

PR #2761, opened 2021-09-27
feat(mwc-button): support associated value

### LINKED ISSUE #2766: [mwc-button] missing value attribute
**Describe the bug**
`mwc-button` should feature `value` attribute and delegate its actual value to shadow tree `button` element to align with [native behavior](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-value).

**Expected behavior**
`mwc-button` submitting a form should define the value associated with the button’s name alongside form data.

**Additional context**
Not sure this should be a bug report or a feature request.
the reason is that button within shadow tree doesn't associate with form in the light tree, therefore supporting value attribute might not even work as expected until browsers support shadow tree form association out of the box.

#2761 sets the ground for this...

yinonov:
add missing `value` attribute as in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-value

Fixes #2766


yinonov:
Relates to #2759

dfreedm:
Thanks for the PR!

packages/button/mwc-button-base.ts
```diff
@@ -108,6 +111,7 @@ export class ButtonBase extends LitElement {
       <button
           id="button"
           class="mdc-button ${classMap(this.getRenderClasses())}"
+          .value="${live(this.value) as unknown as string}"
```
dfreedm, line 114: I don't think the `live` directive is doing much here, as the button value will not change on its own.

Let's remove it!

packages/button/mwc-button-base.ts
```diff
@@ -16,6 +16,7 @@ import {RippleHandlers} from '@material/mwc-ripple/ripple-handlers';
 import {eventOptions, html, LitElement, property, query, queryAsync, state, TemplateResult} from 'lit-element';
 import {ClassInfo, classMap} from 'lit-html/directives/class-map.js';
 import {ifDefined} from 'lit-html/directives/if-defined.js';
+import {live} from 'lit-html/directives/live.js';
```
dfreedm, line 19: Remove this import, since we won't need the `live` directive.

dfreedm (changes requested):

dfreedm:
Oh, and please rebase your PR to use the `lit` package for imports!

packages/button/mwc-button-base.ts
yinonov: sure, keeping `as unknown as string` to align with textfield/textarea. not sure if required here

packages/button/mwc-button-base.ts
```diff
@@ -13,9 +13,17 @@ import '@material/mwc-ripple/mwc-ripple';
...
+import {html, LitElement, TemplateResult} from 'lit';
+import {eventOptions, property, query, queryAsync, state} from 'lit/decorators.js';
+import {ClassInfo, classMap} from 'lit/directives/class-map.js';
+import {ifDefined} from 'lit/directives/if-defined.js';
+=======
 import {eventOptions, html, LitElement, property, query, queryAsync, state, TemplateResult} from 'lit-element';
 import {ClassInfo, classMap} from 'lit-html/directives/class-map.js';
 import {ifDefined} from 'lit-html/directives/if-defined.js';
+import {live} from 'lit-html/directives/live.js';
+>>>>>>> d200836fdaeb9ae4f48853fd6a9c1f068b0e76f8
```
yinonov, line 26:
```suggestion
import {html, LitElement, TemplateResult} from 'lit';
import {eventOptions, property, query, queryAsync, state} from 'lit/decorators.js';
import {ClassInfo, classMap} from 'lit/directives/class-map.js';
import {ifDefined} from 'lit/directives/if-defined.js';
```

packages/button/mwc-button-base.ts
```diff
@@ -108,6 +118,11 @@ export class ButtonBase extends LitElement {
       <button
           id="button"
           class="mdc-button ${classMap(this.getRenderClasses())}"
+<<<<<<< HEAD
+          .value="${this.value as unknown as string}">
+=======
+          .value="${live(this.value) as unknown as string}"
+>>>>>>> d200836fdaeb9ae4f48853fd6a9c1f068b0e76f8
```
yinonov, line 118:
```suggestion
          .value="${this.value as unknown as string}">
```

yinonov:
I purely `git rebase master` and it added all the commits and multiple files changes. Is it normal?

packages/button/mwc-button-base.ts
```diff
@@ -108,6 +111,7 @@ export class ButtonBase extends LitElement {
       <button
           id="button"
           class="mdc-button ${classMap(this.getRenderClasses())}"
+          .value="${this.value as unknown as string}">
```
yinonov, line 114:
```suggestion
          .value="${this.value as unknown as string}"
```

yinonov:
same change with cleaner commit history #2802. can close this 1

PR #2802, opened 2021-10-03
feat(mwc-button): support associated value

yinonov:


yinonov:
Closes #2761

asyncLiz:
Closing this as obsolete. We don't currently plan on supporting button values for form submission.

PR #2814, opened 2021-10-07
Migrate to new "lit" package

bsidhom:
This changes usages of `lit-element` and `lit-html` to the new `lit` npm package. Before this change, downstream Typescript users get a deprecation warning due to the use of `lit-element` in CSS type declarations.

I went ahead and migrated demo and test scripts as well, though I don't think those impact users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ node_modules
 /build
 /coverage
 .DS_Store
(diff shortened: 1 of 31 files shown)


dfreedm:
Thanks for the PR!

PR #2823, opened 2021-10-10
chore(scripts): styles template align with lit 2

yinonov:


allan-chen:
Thanks for the PR! Looks like this has already been fixed.

PR #2832, opened 2021-10-11
Update sass-to-lit-css script to use lit import instead of lit-element

aomarks:


allan-chen (approved):
Thanks! Pulled in for internal review

PR #2853, opened 2021-10-16
fix(mwc-icon-button) Fix alignment when slot is used

bramkragten:
Since flex is now used for alignment, the `i` element, that is always rendered even when not used, now messes the alignment up when the `slot` is used.

image

This PR fixes that by not rendering the `i` element when it is not used.


dfreedm:
@bramkragten Thanks for the PR!

I had to fix some internal screenshot tests, so there's a duplicate PR generated by our system that will resolve when landed internally

PR #2896, opened 2021-11-02
fix(button): support native attributes

yinonov:


components/button/lib/button.ts
```diff
@@ -30,6 +30,10 @@ export abstract class Button extends LitElement {
 
   @property({type: Boolean, reflect: true}) disabled = false;
 
+  @property({type: String}) type: 'button' | 'submit' | 'reset' = 'submit';
+
+  @property({type: String}) value = '';
```
dfreedm, line 35: This will need to implement the FormDataEvent hooks similar to https://github.com/material-components/material-web/blob/master/packages/base/form-element.ts#L61-L103 to be actually sent to the server

dfreedm (changes requested):

asyncLiz:
Closing for now since we don't currently plan on supporting button values in form submission.

PR #2958, opened 2021-12-07
Adding part attribute to elements

ainyava:
Hi
I do understand that the reason of using Web Components is isolating the components but:

### Problem
We were using this library and facing a main problem: **Customization**
So basically we needed to Fork/Extend this repository and manipulate the `SASS` codes to do our customization.

but even then when we need small change like changing a `padding` property or do some `RTL` stuff, the **Shadow Root** does not allow us to do that and we are limited to overriding `--mdc-` attributes like `--mdc-typography-font-family` and those features are so limitted.

### Solution
So generally when we use Web Components the best practice to customize outside of it's shadow root is through the [::part (MDN)](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) attribute to access inside element itself.

I added the `part` attribute to all of non web component elements that are used in templates which I found useful.

Sincerely, @ainyava

### DIFF
--- a/packages/button/mwc-button-base.ts
+++ b/packages/button/mwc-button-base.ts
@@ -108,6 +108,7 @@ export class ButtonBase extends LitElement {
     return html`
       <button
           id="button"
+          part="button"
           class="mdc-button ${classMap(this.getRenderClasses())}"
           ?disabled="${this.disabled}"
           aria-label="${this.label || this.icon}"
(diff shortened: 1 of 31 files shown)


allan-chen:
Thanks Hamed for the PR, and I understand the inconvenience of certain properties that need customization not being exposed through CSS variables, resulting the need to fork.

The `::part` attribute is likely not going to be a supported way of customizing Material components in the future. Instead, we'll work on incorporating [design tokens](https://m3.material.io/foundations/design-tokens/overview) into all the components (including layout tokens like padding) to improve customizability as part of the M3 refresh, where each token corresponds to a CSS custom property. This ensures we provide the maximum customizability without being too open-ended with the API.

If there's one-off RTL bugs unrelated to customization that need addressing, feel free to also open an issue so we can update the base stylesheet to fix those bugs.

Thanks!

PR #3057, opened 2022-01-13
chore: remove hardcoded lit/reactive-element version

nelsonsilva:
In https://github.com/material-components/material-web/commit/6278ee5df98ca7eae0d694e58b052fdbf4b0519d we aligned on lit^2.0.0 but kept an explicit version of @lit/reactive-element for no apparent good reason. This just cleans it up

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -11,7 +11,6 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


nelsonsilva:
Just rebased on master and fixed the package.json conflict @asyncLiz

PR #3086, opened 2022-01-28
Add ARIA none role to mwc-ripple

steverep:


steverep:
Oops, reformatted the whole file by mistake 🤫

Issue #2986, opened 2021-12-28
ARIA role for mwc-ripple


steverep:
The `mwc-ripple` custom element produces unnecessary nodes in the accessibility tree (one for the custom element and one for the child `div` in the shadow DOM).  Since this component is entirely just meant to be a CSS insertion hook for animation, the element and its child should be given the [ARIA none role](https://www.w3.org/TR/wai-aria-1.1/#none), i.e. add the attribute `role="none presentation"`.

dfreedm:
Makes sense, I'll try to play with that soon and see how it goes.

steverep:
@dfreedm  can you take a look at PR #3087 to close this?

dfreedm:
Yeah sorry, I meant to comment on that yesterday.

I haven't been able to replicate VoiceOver or NVDA announcing anything for the ripple element via object navigation, so I'm pretty sure this change isn't necessary.

MrHBS:
@dfreedm I think the point is to remove it from the tree altogether so you can't even navigate to it.

steverep:
> I haven't been able to replicate VoiceOver or NVDA announcing anything for the ripple element via object navigation, so I'm pretty sure this change isn't necessary.

So, I was going to comment on how to do it, but now I cannot navigate to it either on any of the demo components I've checked.  That was definitely not the case when I filed this, so something must have changed in the past few months either with browsers or OS.

It used to be that after landing on a `mwc-icon-button`, for example, NVDA could navigate all the children, including the empty ripple `div`.  Oddly enough, the ripple element is still in the accessibility tree (see attached).  I don't have a tree from the past to compare, but it looks like it is now being labelled as "offscreen" which I don't remember being the case.

[mwc-icon-button-accessibility-tree.zip](https://github.com/material-components/material-web/files/8291616/mwc-icon-button-accessibility-tree.zip)

In any case, `role="none"` or `aria-hidden="true"` is still the correct and safest treatment here because it's an element that marks up nothing and exists solely to inject CSS.  You could use any block element and get the same effect - the semantic role of the element means nothing.  From the [ARIA 1.2 spec list of example use cases](https://www.w3.org/TR/wai-aria-1.2/#note-regarding-the-aria-1-1-none-role-0):

> An element used as an additional markup "hook" for CSS; or

PR #3087, opened 2022-01-28
Add ARIA presentation role to ripple

steverep:


asyncLiz:
What is this PR solving for? The ripple should not be read by screen readers.

steverep:
> What is this PR solving for? The ripple should not be read by screen readers.

Explained in issue #2986 - just because there is no text for assistive technology to consume doesn't mean it's invisible.  The ARIA none role is exactly for stuff like this.

asyncLiz:
Thanks for the issue link. Can you describe what the behavior of screen readers are when there is not a role attribute on the ripple?

steverep:
> Thanks for the issue link. Can you describe what the behavior of screen readers are when there is not a role attribute on the ripple?

It depends both on the screen reader and browser.  First, users can always navigate the accessibility tree manually when needed (e.g. NVDA calls this object navigation).  The `<div>` just appears as an empty text container or similar (non-semantic elements like `div` or `span` can be in the tree differently depending on the accessibility API).  Adding `role="none"` tells the user agent to ignore the generic role, and since it's empty it effectively removes it from the tree so it won't be navigated in that mode.

Second, user agents are supposed to treat all children of things such as buttons as presentation/none, but that doesn't always happen, especially when shadow boundaries are involved.  When they aren't, it can cause confusion when browsed because the boundaries of such elements are not announced and semantics inside them are not conventional/expected.

Adding the role as I show here is the right thing to do.  I would encourage you to read about these topics in the W3C ARIA 1.2 spec and related documents for greater detail.

steverep:
I'm willing to fix other accessibility issues, but is this project not reviewing outside pull requests?  This is not a particularly taxing review at 1 changed line.

asyncLiz:
At the moment 100% of our energy is focused towards Material You, which is why only critical PRs are currently being reviewed.

We've gone through our internal a11y review and have determined that it is not necessary to add this role. If a screenreader is announcing arbitrary `<div>`s, I'd recommend creating a small repro and filing a bug on that screenreader!

steverep:
> We've gone through our internal a11y review and have determined that it is not necessary to add this role. If a screenreader is announcing arbitrary <div>s, I'd recommend creating a small repro and filing a bug on that screenreader!

There's no screen reader problem here, and nothing I've commented has suggested that the issue was with arbitrary announcements.

I've acknowledged on the issue thread that the behavior has changed since I first filed it, but I've also commented in detail and pointed to W3C references as to why this ought to be removed from the accessibility tree.  The latter has largely gone without response.

PR #3174, opened 2022-02-19
Add fullwidth property to Properties section

omar2205:
Add the missing `fullwidth` property


asyncLiz (approved):
Thanks for adding this!

PR #3197, opened 2022-02-24
fix(icon-button): update action-element path

mylmz10:


asyncLiz:
@dfreedm can you update copybara for this?

PR #3351, opened 2022-04-19
[tests] remove saucelabs from karma config and github actions

taylor-vann:
Remove saucelabs from material-web karma config and github options

it is not needed or used but they run!


asyncLiz (approved):
LGTM

taylor-vann:
thank youuuu

PR #3398, opened 2022-05-05
Add support for aria-expanded to button and icon button

### LINKED ISSUE #3245: Buttons do not support aria-expanded
**Describe the bug**
The APIs for `mwc-button` and `mwc-icon-button` have no way to set the `aria-expanded` attribute on the `<button>` element.  This is needed, for example, when the button is used to open a popup menu.  It should be set to true or false based on user input, and omitted when null or the empty string.

**To Reproduce**
n/a

**Expected behavior**
Button components should support all potential ARIA attributes as defined in the W3C spec and transfer those attributes and states to the internal `<button>` element.

**Screenshots**
n/a

n/a

**Additional context**
[ARIA 1.2 Authoring Practice for menu buttons](https://www.w3.org/TR/wai-aria-practices/#menubutton)

steverep:
Fixes #3245 by adding support for the `aria-expanded` attribute in the same manner as `aria-haspopup`.

### DIFF
--- a/packages/button/README.md
+++ b/packages/button/README.md
@@ -123,6 +123,7 @@ mwc-button {
 | Name | Type | Default | Description
 | ---- | ---- | ------- | -----------
 | `aria-haspopup` | `string` | `undefined` | Indicates the availability and type of an interactive popup element, such as menu or dialog, that can be triggered by the button.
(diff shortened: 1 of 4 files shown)


asyncLiz:
Closing this in favor of #4026

PR #3439, opened 2022-05-16
Add focus method to tab bar

### LINKED ISSUE #3396: Cannot call focus on tab bar
**Describe the bug**
The `mwc-tab-bar` component cannot be focused programmatically via focus method.  For example, if a tab bar is desired to gain focus inside a dialog, focusing the active tab by calling focus on the tab bar element is not possible.

**To Reproduce**
1. Try to call `focus` on `mwc-tab-bar` and nothing happens

**Expected behavior**
Focus should be placed on the active tab.

 - OS: Windows 10
 - Browser Firefox or Chrome
 - Version latest

**Additional context**
[ARIA Authoring Practices on Tabs](https://www.w3.org/TR/wai-aria-practices-1.2/#tabpanel)

steverep:
Fixes #3396 by adding a simple focus method to focus the currently active tab index.

### DIFF
--- a/packages/tab-bar/README.md
+++ b/packages/tab-bar/README.md
@@ -263,6 +263,7 @@ _Note: example is in the state of hovering over the first tab._
 
 | Name     | Description
 | -------- | -------------
+| `focus() => void` | Focuses the currently active tab.
 | `scrollIndexIntoView(index:number) => void` | For long, scrollable `tab-bar`s, scrolls the tab at the given index into view.
 
 ### Events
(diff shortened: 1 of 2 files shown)


steverep:
Should I be modifying the equivalent file in the components folder? It's not immediately clear to me why the code is repeated in both places.

steverep:
I gather from the commit log that the packages directory is probably being deprecated in favor of the components directory.  Is that correct?  If so, are there instructions on how to test and run those demos?

I know the team is working on Material 3 upgrades.  I'd like to submit some accessibility fixes for review so they are available ASAP.  Let me know how best to contribute.  Thanks.

asyncLiz:
Obsolete with M3

PR #3448, opened 2022-05-23
docs: update mwc checkbox to the correct package

gstvribs:


asyncLiz:
This is obsolete now with the current readme

PR #3734, opened 2022-10-22
Add "type": "module" to package.json

### LINKED ISSUE #3732: "type": "module" is missing from package.json
This means that imports do not work in Node, which will interpret files as scripts instead of modules.

justinfagnani:
This is needed to load modules in Node for SSR.

Fixes #3732

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@material/web",
   "version": "0.1.0-alpha.0",
+  "type": "module",
   "publishConfig": {
     "access": "public"
   },


asyncLiz (approved):
LGTM to import

fetishcoder:
any update on this?

asyncLiz:
[Fixed now](https://github.com/material-components/material-web/blob/07207b43aec9d27173902a762d6bdd89c4e7a6dc/package.json#L24)

PR #3799, opened 2022-11-30
Set up existing Jasmine tests on GitHub CI

aomarks:
- The existing Jasmine tests can now be run with `npm test`. It uses https://modern-web.dev/docs/test-runner/overview/ with the https://github.com/blueprintui/web-test-runner-jasmine plugin. **They all pass on Chromium but not Firefox, and I didn't try Safari yet, so I'm only including Chromium for now.**

- The only incompatibility with the test files are the use of `import 'jasmine'`, which is done in Google3 to make `taze` happy, but is invalid externally because the `jasmine` package is a Node-only commonjs module that can't be loaded in the browser (https://unpkg.com/browse/jasmine@4.5.0/lib/jasmine.js), so the test runner jasmine plugin is responsible for actually injecting the browser-compatible Jasmine. The copybara transform has been updated in https://github.com/material-components/material-web/commit/933cb14ac8ee02b0d3a8b36ef8045472adecfb32.

- Added `build` and `test` GitHub actions, which should both pass. The reason I separated them is because it makes the CI status easier to read in the case that the build succeeds but the tests fail.

- The `package-lock.json` file is now checked in. I think we should start doing this, because it makes CI builds more reliable and much easier to know when a change in a dependency is what broke the build/tests, since those upgrades are tracked by Git.

- Converted the package to `"type": "module"`, which is important to help tools understand that this package distributes modules, not commonjs which is the default. I also updated `css-to-ts.js` to be a module since it is now executed as one by default (trivial change).

- Upgraded dependencies.

Part of https://github.com/material-components/material-web/issues/3773


asyncLiz:
I have concerns about going back to google3 and github-specific files. In MDC and MWC this commonly resulted in the external github files becoming stale since the project is google3-first.

What are the downsides to checking in all files and keeping google3 and github perfectly in sync? Maybe we can make a one-pager to discuss?

aomarks:
> I have concerns about going back to google3 and github-specific files. In MDC and MWC this commonly resulted in the external github files becoming stale since the project is google3-first.
> 
> What are the downsides to checking in all files and keeping google3 and github perfectly in sync? Maybe we can make a one-pager to discuss?

We discussed offline, and @asyncLiz convinced me that there are some disadvantages to not tracking all files in google3, and the advantages I was thinking of are not so significant. So I retract this suggestion.

aomarks:
To add some detail to the `// import 'jasmine';` issue, it's not just that `jasmine` externally is commonjs (which could be resolved with a rollup commonjs transform), it's that `jasmine` externally is a Node-only module which imports `os`, `path`, etc: https://unpkg.com/browse/jasmine@4.5.0/lib/jasmine.js

So Google3 has the `jasmine` specifier configured differently to how it actually works externally.

aomarks:
Merged with master which has the copybara transforms (https://github.com/material-components/material-web/commit/933cb14ac8ee02b0d3a8b36ef8045472adecfb32), so I think this is ready to land.

dfreedm (approved):
LGTM.

What sort of issues are happening on firefox?

web-test-runner.config.js
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {playwrightLauncher} from '@web/test-runner-playwright';
+import {jasmineTestRunnerConfig} from 'web-test-runner-jasmine';
+
+export default {
+  ...jasmineTestRunnerConfig(),
+  nodeResolve: true,
+  files: ['**/*_test.js', '!node_modules/', '!.wireit/'],
```
asyncLiz, line 13: Optional: thoughts on matching `*test.js` instead? 

We have some mixed `file_test.ts` and `file.test.ts` naming conventions that need to be resolved (although the `*.test.ts` ones aren't exported, so this will be fine for now)

asyncLiz (approved):

web-test-runner.config.js
aomarks: Done

aomarks:
> What sort of issues are happening on firefox?

Firefox fails with "Touch is not defined" on 11 tests (which seems weird, but indeed a Touch global is not defined when I check Firefox in the console, even though MDN says the Touch constructor is supported).

Safari just fails with "unknown error".

Added comments.

dfreedm:
Firefox doesn't define Touch unless it detects a touchscreen. I don't think we check for Touch anywhere in MWC, so it must be a web-test-runner thing...

aomarks:
> Firefox doesn't define Touch unless it detects a touchscreen. I don't think we check for Touch anywhere in MWC, so it must be a web-test-runner thing...

I think it might be this line: https://github.com/material-components/material-web/blob/933cb14ac8ee02b0d3a8b36ef8045472adecfb32/testing/harness.ts#L728

dfreedm:
Ah that makes sense

PR #4026, opened 2023-03-06
feat(button,iconbutton): add support for aria-expanded to buttons

### LINKED ISSUE #3245: Buttons do not support aria-expanded

steverep:
Fixes #3245 by adding support for the `aria-expanded` attribute in the same manner as `aria-haspopup`.

Note this is identical to #3398 minus the docs since they don't exist yet.

### DIFF
--- a/button/lib/button.ts
+++ b/button/lib/button.ts
@@ -20,7 +20,7 @@ import {ariaProperty} from '../../decorators/aria-property.js';
 import {pointerPress, shouldShowStrongFocus} from '../../focus/strong-focus.js';
 import {ripple} from '../../ripple/directive.js';
 import {MdRipple} from '../../ripple/ripple.js';
-import {ARIAHasPopup} from '../../types/aria.js';
+import {ARIAExpanded, ARIAHasPopup} from '../../types/aria.js';
 
 import {ButtonState} from './state.js';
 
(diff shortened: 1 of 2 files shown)


button/lib/button.ts
```diff
@@ -29,6 +29,10 @@ export abstract class Button extends LitElement implements ButtonState {
   static override shadowRootOptions:
       ShadowRootInit = {mode: 'open', delegatesFocus: true};
 
+  @property({type: String, attribute: 'data-aria-expanded', noAccessor: true})
+  @ariaProperty
+  override ariaExpanded!: ARIAExpanded;
+
   @property({type: String, attribute: 'data-aria-has-popup', noAccessor: true})
```
steverep, line 36: Unrelated, but this seems like an error to me; `haspopup` should not be hyphenated.

steverep:
@asyncLiz is the merge delay normal here or is the failed check the problem?  If so, I think just a rerun would fix it?

asyncLiz:
Thank you for the ping! Looks like I missed the email for importing it. We'll run some checks, but LGTM!

PR #4111, opened 2023-04-01
chore: update environment.ts

MrHBS:
I believe this should work too, right @asyncLiz ?


asyncLiz:
Yes, this should work too.

> Also @asyncLiz this file keeps mentioning about it being an environment for screenshot tests

There are, they're just internal only since external screenshot testing is too flaky 🫠

asyncLiz (approved):

PR #4284, opened 2023-05-12
docs(catalog): add catalog

e111077:


e111077:
Catalog has been merged into main and should be plugged into GH actions with #4382. Branch is now obsolete

PR #4295, opened 2023-05-16
Fix wireit taskname copy-readmes

vdegenne:


asyncLiz:
@e111077 can you review this?

e111077 (approved):
Nice catch, thanks

PR #4306, opened 2023-05-17
build: update package.json

vdegenne:
On my way learning how eleventy works I found out it has a default cjs configuration file name it's looking for.
I just PR to share, if you guys don't like micro PRs like that, please tell me, I do not mean to be bothering, just sharing things as I learn.

note: this filename was added to the default filenames list > 2.0.0 (https://www.11ty.dev/docs/config/#default-filenames)

*UPDATE:* I also removed a npm run inside a wireit command that is already in the dependencies of the same task.


vdegenne:
Requesting your review : @e111077

e111077:
thanks!

e111077:
apologies, this is a WIP branch and there are lots of rebases happening for quick iteration. It might not be merged but instead merged from the google-side

vdegenne:
@e111077 I think I understand (I'm not familiar with git that much), by "rebase" you mean you did integrate the new changes from main to this branch, right?
That's why when I committed a change it showed the commits in between?

e111077:
If i did this the normal way, i should have merged the new changes from the main branch into this one which would make the commit tree look like it merged back into mine after it diverged. Instead i rebased which means i took the initial commit that branches from the main branch, and just moved it to the tip commit of the main branch.

By rebasing, i have to force push since i've rewritten git history, hence what broke you

The hero image of this medium article shows the difference

https://medium.com/@basecs101/do-you-know-the-difference-between-git-merge-and-git-rebase-latest-800983738b60

vdegenne:
Thanks a lot for the explanation. Makes perfect sense now! :deciduous_tree:

PR #4311, opened 2023-05-18
build: setup the npm workspace for the catalog

e111077:
We have to do this externally because we don't sync the package-lock internally and CI runs `npm ci` which breaks due to a mismatch between package.json and package-lock.json


asyncLiz (approved):
lgtm

e111077:
OMG COPYBARA COPY!

PR #4314, opened 2023-05-21
docs: fix button labels

kattschan:
The current docs provide a label property that does not work, so this doc changes that to text inside the element


asyncLiz (approved):
Thanks for the fix!

PR #4337, opened 2023-05-25
fix(list): fix focus ring shown on list-item-only

VandeurenGlenn:


asyncLiz (approved):
Good catch!

list/lib/listitem/list-item-only.ts
```diff
@@ -26,4 +26,8 @@ export class ListItemOnly extends ListItemEl {
   override renderRipple() {
     return this.noninteractive ? nothing : super.renderRipple();
   }
+
```
asyncLiz, line 29: small nit: can you remove the trailing whitespace on this line?

asyncLiz (changes requested):

list/lib/listitem/list-item-only.ts
VandeurenGlenn: Sure!

VandeurenGlenn: @asyncLiz  I think that did it

asyncLiz: Thanks!

list/lib/listitem/list-item-only.ts
```diff
@@ -26,4 +26,8 @@ export class ListItemOnly extends ListItemEl {
...
+  override renderFocusRing() {
```
asyncLiz, line 30: Uh oh, build errors! We need to add an explicit type to the `super.renderFocusRing()`, just like [`renderRipple()` does](https://github.com/material-components/material-web/blob/33daf19c73ae1599a10286dada80edf288b52011/list/lib/listitem/list-item.ts#L144).

Update `list-item.ts` and change that method return type to

```
protected renderFocusRing(): TemplateResult|typeof nothing {
```

asyncLiz (changes requested):

list/lib/listitem/list-item-only.ts
VandeurenGlenn: Aha, Think it's best I just open it up in vs code instead of editing on GH 😅

vdegenne: @VandeurenGlenn it's just one line :eyes:

VandeurenGlenn: Yes well, @vdegenne the rest wasn't also much but still gave an error :p

vdegenne: @VandeurenGlenn well you messed up :rofl:

asyncLiz: iono what happened, but our internal copy tool did NOT like that 😂

VandeurenGlenn: @asyncLiz Done, also made noninteractive not clickable

list/lib/listitem/_list-item.scss
```diff
@@ -73,11 +73,11 @@
     // hide android tap color since we have ripple
     -webkit-tap-highlight-color: transparent;
 
-    &:not(.disabled):not(.noninteractive) {
+    &:not(.disabled) {
       cursor: pointer;
     }
 
-    &.disabled {
+    &.disabled, &.noninteractive {
```
asyncLiz, line 80: What is the use case for disabling pointer events when non-interactive? I would like to still give users the option to add pointer event handlers for the own interactions, especially since you could just set `pointer-events: none` on the host with CSS if that's the desired effect

VandeurenGlenn: @asyncLiz It's all @vdegenne his fault 👀 
No, i think, if it's non-interactive it makes sense it isn't clickable because you wouldn't be shown a ripple when its clicked plus I think it also shouldn't be intractable with the keyboard

asyncLiz: I think there are use cases for a ripple that does not have a ripple/focus-ring but can still allow users to listen to click events and perform actions.

At the very least, allowing pointer events gives that option, which an easy workaround to disable them entirely via CSS for those who need them prevented.

VandeurenGlenn: Okido, I'll force push again :p

list/lib/listitem/list-item-only.ts
```diff
@@ -12,7 +12,7 @@ import {ListItemEl} from './list-item.js';
 // tslint:disable-next-line:enforce-comments-on-exported-symbols
 export class ListItemOnly extends ListItemEl {
   /**
-   * Removes the hover and click ripples from the item when true.
+   * Removes the hover and click ripples from the item & sets pointer-events to none when true.
```
e111077, line 15: please revert comment

e111077 (changes requested):

e111077:
Also need to merge main in. There seems to be a conflict somewhere causing issues

list/lib/listitem/list-item-only.ts
VandeurenGlenn: ah woops

VandeurenGlenn:
@e111077 Before is saw same unmergeable but I think it's fine now or can you share the conflicts to resolve?

e111077:
ah nevermind must have been an old email!

e111077 (approved):

PR #4350, opened 2023-05-26
fix(list): infer noninteractive property type

vdegenne:
- Adding `type: Boolean` to `noninteractive` property.
- Prevent focus ring to appear when `noninteractive` is set to `true`

### DIFF
--- a/list/lib/listitemlink/list-item-link-only.ts
+++ b/list/lib/listitemlink/list-item-link-only.ts
@@ -15,7 +15,7 @@ export class ListItemLinkOnly extends ListItemLink {
    * Removes the hover and click ripples from the item when true. Clicking the
    * link will still cause link navigation.
    */
-  @property() noninteractive = false;
+  @property({type: Boolean}) noninteractive = false;
 
   override getRenderClasses() {
     return {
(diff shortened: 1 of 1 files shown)


list/lib/listitemlink/list-item-link-only.ts
```diff
@@ -27,4 +27,8 @@ export class ListItemLinkOnly extends ListItemLink {
   override renderRipple() {
     return this.noninteractive ? nothing : super.renderRipple();
   }
+
+  override renderFocusRing() {
```
asyncLiz, line 31: With https://github.com/material-components/material-web/pull/4337 in flight, can we limit the scope of this PR to just adding `{type: Boolean}`?

asyncLiz (changes requested):

list/lib/listitemlink/list-item-link-only.ts
vdegenne: @asyncLiz not same file :eyes:

vdegenne: @asyncLiz I can if you want remove `renderFocusRing` if you prefer having this change in #4333 but can @VandeurenGlenn really do it without messing things is the question :eyes:

VandeurenGlenn: @vdegenne I was doing to much at once ok 😂

vdegenne:
@e111077 Can I request your review?
I saw you added `renderFocusRing` in another PR. So this PR is just going to add Boolean type hint to the `noninteractive` property.

PR #4378, opened 2023-05-31
chore(dialog): provide transition values type

vdegenne:
This PR provides the `transition` values type.

This would help with type checking when using `MdDialog` programmatically and also to forward types in tool authoring, e.g.

```typescript
export interface DialogOptions {
	// ...

	/**
	 * Transition of the dialog.
	 * @type {import('@material/web/dialog/dialog.js').MdDialog['transition']}
	 * @default 'grow-down'
	 */
	transition?: MdDialog['transition'];
}
```

### DIFF
--- a/dialog/demo/demo.ts
+++ b/dialog/demo/demo.ts
@@ -15,9 +15,9 @@ import { DialogTransition } from '../dialog.js';
 
 const collection =
     new MaterialCollection<KnobTypesToKnobs<StoryKnobs>>('Dialog', [
-      new Knob<Extract<DialogTransition, string>, "transition">('transition', {
-        defaultValue: 'grow-down',
-        ui: selectDropdown({
+      new Knob('transition', {
+        defaultValue: 'grow-down' as DialogTransition,
+        ui: selectDropdown<DialogTransition>({
           options: [
             {label: 'grow-down', value: 'grow-down'},
             {label: 'grow-up', value: 'grow-up'},
--- a/dialog/demo/demo.ts
+++ b/dialog/demo/demo.ts
@@ -11,10 +11,11 @@ import {KnobTypesToKnobs, MaterialCollection, materialInitsToStoryInits, setUpDe
 import {boolInput, Knob, selectDropdown, textInput} from './index.js';
 
 import {stories, StoryKnobs} from './stories.js';
+import { DialogTransition } from '../dialog.js';
 
 const collection =
     new MaterialCollection<KnobTypesToKnobs<StoryKnobs>>('Dialog', [
-      new Knob('transition', {
+      new Knob<Extract<DialogTransition, string>, "transition">('transition', {
         defaultValue: 'grow-down',
         ui: selectDropdown({
           options: [
--- a/dialog/lib/dialog.ts
+++ b/dialog/lib/dialog.ts
@@ -26,7 +26,7 @@ const CLOSING_TRANSITION_PROP = '--_closing-transition-duration';
 /**
  * Dialog transition type.
  */
-export type DialogTransition = 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right';
+export type DialogTransition = 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right'|'';
 
 /**
  * A dialog component.
--- a/dialog/lib/dialog.ts
+++ b/dialog/lib/dialog.ts
@@ -23,6 +23,9 @@ export const CLOSE_ACTION = 'close';
 const OPENING_TRANSITION_PROP = '--_opening-transition-duration';
 const CLOSING_TRANSITION_PROP = '--_closing-transition-duration';
 
+/**
+ * Dialog transition type.
+ */
 export type DialogTransition = 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right';
 
 /**
--- a/dialog/lib/dialog.ts
+++ b/dialog/lib/dialog.ts
@@ -203,7 +203,8 @@ export class Dialog extends LitElement {
    *
    * Defaults to grow-down.
    */
-  @property({reflect: true}) transition = 'grow-down';
+  @property({reflect: true})
+  transition: 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right' = 'grow-down';
 
   private currentAction: string|undefined;
 
--- a/dialog/lib/dialog.ts
+++ b/dialog/lib/dialog.ts
@@ -23,6 +23,8 @@ export const CLOSE_ACTION = 'close';
 const OPENING_TRANSITION_PROP = '--_opening-transition-duration';
 const CLOSING_TRANSITION_PROP = '--_closing-transition-duration';
 
+export type DialogTransition = 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right';
+
 /**
  * A dialog component.
  */
@@ -203,8 +205,7 @@ export class Dialog extends LitElement {
    *
    * Defaults to grow-down.
    */
-  @property({reflect: true})
-  transition: 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right' = 'grow-down';
+  @property({reflect: true}) transition: DialogTransition = 'grow-down';
 
   private currentAction: string|undefined;
 
--- a/dialog/demo/stories.ts
+++ b/dialog/demo/stories.ts
@@ -14,12 +14,12 @@ import '@material/web/button/text-button.js';
 import '@material/web/dialog/dialog.js';
 
 import {MaterialStoryInit} from './material-collection.js';
-import {MdDialog} from '@material/web/dialog/dialog.js';
+import {MdDialog, DialogTransition} from '@material/web/dialog/dialog.js';
 import {css, html} from 'lit';
 
 /** Knob types for dialog stories. */
 export interface StoryKnobs {
-  transition: string|undefined;
+  transition?: DialogTransition;
   fullscreen: boolean;
   modeless: boolean;
   footerHidden: boolean;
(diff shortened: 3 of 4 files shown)


dialog/lib/dialog.ts
```diff
@@ -203,7 +203,8 @@ export class Dialog extends LitElement {
    *
    * Defaults to grow-down.
    */
-  @property({reflect: true}) transition = 'grow-down';
+  @property({reflect: true})
+  transition: 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right' = 'grow-down';
```
asyncLiz, line 207: Can you make this a type at the top of the file?

```ts
export type DialogTransition = 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right';
```

Then re-export it in `dialog/dialog.ts` for easier use by clients.

```ts
export {DialogTransition} from './lib/dialog';
```

asyncLiz (changes requested):

dialog/lib/dialog.ts
vdegenne: Ok, done. Wdyt?

asyncLiz (approved):
LGTM

dialog/lib/dialog.ts
```diff
@@ -23,6 +23,8 @@ export const CLOSE_ACTION = 'close';
 const OPENING_TRANSITION_PROP = '--_opening-transition-duration';
 const CLOSING_TRANSITION_PROP = '--_closing-transition-duration';
 
+export type DialogTransition = 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right';
```
asyncLiz, line 26: oh oof, we need jsdoc for this, just realized. Mind adding some? It can just say "Dialog transition type."

asyncLiz (changes requested):
LGTM

vdegenne:
Another suggestion before you merge this PR. should it not be 
```typescript
export type {DialogTransition} from './lib/dialog';
```
?

asyncLiz (changes requested):
Can you add the type to the [`demo/stories.ts`](https://github.com/material-components/material-web/blob/dde57a178d99133ac41983a4353d204036c167c6/dialog/demo/stories.ts#L22) file?

vdegenne:
Alright :+1:

vdegenne:
> Can you add the type to the [`demo/stories.ts`](https://github.com/material-components/material-web/blob/dde57a178d99133ac41983a4353d204036c167c6/dialog/demo/stories.ts#L22) file?

Ok it should be good.

vdegenne:
By the way, out of the scope of this PR but why you guys don't use `StoryKnobs` interface for the parameters?

https://github.com/material-components/material-web/blob/dde57a178d99133ac41983a4353d204036c167c6/dialog/demo/stories.ts#L40-L50

asyncLiz:
I believe typescript infers those types automatically, so there's no need to explicitly type them :)

asyncLiz (approved):

asyncLiz (changes requested):
Getting a new error in the stories:
```
dialog/demo/stories.ts:115:18 - error TS2322: [lit] Type '"condense"' is not assignable to 'DialogTransition'

115                            ~~~~~~~~~~
```

I actually don't know what the condense transition is supposed to be. Maybe swap it with `shrink`?

vdegenne:
Couldn't find the `condense` transition in the specs :eyes:

asyncLiz (changes requested):
absdoigjs more build errors 🙃

```
dialog/demo/demo.ts:17:7 - error TS2322: Type 'Knob<string | undefined, "transition">' is not assignable to type 'Knob<DialogTransition | undefined, "transition"> | Knob<boolean, "fullscreen"> | Knob<boolean, "modeless"> | Knob<boolean, "footerHidden"> | ... 4 more ... | Knob<...>'.
  Type 'Knob<string | undefined, "transition">' is not assignable to type 'Knob<DialogTransition | undefined, "transition">'.
    Type 'string | undefined' is not assignable to type 'DialogTransition | undefined'.
      Type 'string' is not assignable to type 'DialogTransition | undefined'.

 17       new Knob('transition', {
          ~~~~~~~~~~~~~~~~~~~~~~~~
 18         defaultValue: 'grow-down',
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 29         })
    ~~~~~~~~~~
 30       }),
    ~~~~~~~~
```

Can you try changing it to `defaultValue: 'grow-down' as DialogTransition,`?

dialog/demo/demo.ts
```diff
@@ -11,10 +11,11 @@ import {KnobTypesToKnobs, MaterialCollection, materialInitsToStoryInits, setUpDe
 import {boolInput, Knob, selectDropdown, textInput} from './index.js';
 
 import {stories, StoryKnobs} from './stories.js';
+import { DialogTransition } from '../dialog.js';
 
 const collection =
     new MaterialCollection<KnobTypesToKnobs<StoryKnobs>>('Dialog', [
-      new Knob('transition', {
+      new Knob<Extract<DialogTransition, string>, "transition">('transition', {
```
vdegenne, line 18: Can you check that out @asyncLiz @e111077 ?

That was the only way I could get rid of the error because `KnobTypesToKnobs` *extracts* the values of the generic so `DialogTransition` was no longer an option.
Maybe it would be better to change `KnobTypesToKnobs` directly? (I'm not a TypeScript wizard :disappointed_relieved: )

dialog/lib/dialog.ts
```diff
@@ -26,7 +26,7 @@ const CLOSING_TRANSITION_PROP = '--_closing-transition-duration';
 /**
  * Dialog transition type.
  */
-export type DialogTransition = 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right';
+export type DialogTransition = 'grow'|'shrink'|'grow-down'|'grow-up'|'grow-left'|'grow-right'|'';
```
vdegenne, line 29: Also I thought good to add the empty option to the list, that is also one value in the knob select drop down.

dialog/demo/demo.ts
vdegenne: I just confirm that this explicit generic makes TypeScript use types in proper places. It just doesn't feel right having to explicitly specify that.

e111077: The thing is internally we actually do hook it up to the build system, but externally, even in the catalog, it is run in the playground which does not run any build processes. I might have to consider hooking this into the extenal build system as well. In the meantime I think there are 2 options:

1. follow the typing examples in [fab/demo/demo.ts](https://github.com/material-components/material-web/blob/main/fab/demo/demo.ts#L21-L29) and pass the generics and cast the default value
2. just skip the stories part and we'll do this for you since it just might be easier to do that ourselves

vdegenne: @e111077 Ok I just updated , I understand now.

e111077: A bit of context, is that Lit stories internally is framework agnostic (lit / angular / whatever) and is meant to be in a single file where the types are inferred from the `Knob()` array. We wanted to externalize these stories but only expose the simplest of stuff we could to the viewers of the catalog, so we had to do KnobTypesToKnobs shenanigans to format it in a way such that stories.ts is simple and the "source of truth", but still integrates with the internal build of our Lit stories system.

This is why there is some weirdness between that type and the `new Knob()` array.

e111077: TBH I think this requires a rewrite of `selectDropdown`, but so many projects at google use lit stories, that i've been putting off that migration

vdegenne: @e111077 Thanks a lot for the explanation, it means a lot. Definitely not easy to grasp some concept because I'm not familiar with the framework plus that stories TypeScript is crazy. But I see how generics work now.

I actually have a bigger question, what's your TS workflow for the demos? I manually copied the content of `stories` from the catalog to the dialog `demo` directory because the demo file was trying to import these files relatively..

vdegenne: @e111077 Also could it be something is wrong with TypeScript from too much complexity? `KnobTypesToKnobs<StoryKnobs>` gives a decent list of knob types but somehow `DialogTransition` is lost in the `MaterialCollection` class template...
So writing
```typescript
new Knob<DialogTransition, "transition">('...
```
back get rid of the error.

asyncLiz:
Hmm, I feel like this might be better to copy and author internally so we can catch all the side effects and skip the back and forth churn. Does that sound good?

dialog/demo/demo.ts
e111077: I agree that there are too many complex types, but that mostly stems from the complex types inside of Stories itself, which, in order to keep the same interface between internal and external build, might be difficult to simplify.

I feel the `KnobTypesToKnobs` complexity mostly stems from the complexity of stories itself and the attempt to keep it typesafe between stories.ts and demo.ts in each demo.

If you have any suggestions to change KnobTypesToKnobs to be simpler, i'd be very interested, but modifying the types inside Stories itself e.g. `Knob`, `Collection`, `LitCollection`, or `PolymorphicArrayOfKnobs`, there might be much more to consider

vdegenne: I think the `KnobTypesToKnobs` type is fine, but for some reason I can't find a solution to that problem, the problem is that if there are string type values in the story the type string will get mixed with string union types (`DialogTransition` in this case) which causes the type to be lost in translation. I made a TypeScript playground : https://tsplay.dev/NBQMnw
 Maybe that can help you to find a solution?

vdegenne: oh and nevermind, this will actually help intellisense but still getting some errors.
The only suggestion I would make is to export the `KnobInit` type so we can write:

```typescript
new Knob('transition', {
  defaultValue: 'grow-down',
  ui: selectDropdown({
    options: [
      {label: 'grow-down', value: 'grow-down'},
      // ...
    ]
  })
} as KnobInit<DialogTransition>),
```
only to avoid a double cast. (or else it's fine)

asyncLiz:
Obsolete with https://github.com/material-components/material-web/pull/4677

PR #4382, opened 2023-05-31
build(ci/catalog): gh actions to build & delploy catalog

e111077:
~Requires #4380 and #4381 to merge before it builds~

Fixes b/285197169


asyncLiz:
lgtm to import

PR #4430, opened 2023-06-08
Update quick-start.md (fix playground example)

VandeurenGlenn:


asyncLiz:
Thanks for pointing this out! I've updated the original playground, so no need to change the gist :)

PR #4496, opened 2023-06-24
chore(dialog): transition improvements code

vdegenne:
Few improvements:

- fix a typo in `_dialog.scss` (`trasition` => `transition`)
- condition `performTransition` to only perform a transition when `this.open` new and old values are different.
- set opacity to 0 on `containerElement` when it's closed and 1 when it opens, so opening and closing animation feels smoother.

### DIFF
--- a/dialog/lib/_dialog.scss
+++ b/dialog/lib/_dialog.scss
@@ -194,7 +194,9 @@
   // Transitions for open/closed states
   .container {
     will-change: transform, opacity;
+    // TODO: make opacity fade out content and container separately
     transition-property: transform, opacity;
+    transition-timing-function: var(--_closing-transition-easing), linear;
     overflow: hidden;
   }
 
--- a/dialog/lib/_dialog.scss
+++ b/dialog/lib/_dialog.scss
@@ -110,6 +110,8 @@
     max-inline-size: var(--_container-max-inline-size);
     padding-block-start: var(--_container-block-padding);
     padding-block-end: var(--_container-block-padding);
+
+    opacity: 0;
   }
 
   md-elevation {
@@ -192,7 +194,7 @@
   // Transitions for open/closed states
   .container {
     will-change: transform, opacity;
-    transition-property: transform;
+    transition-property: transform, opacity;
     overflow: hidden;
   }
 
--- a/dialog/lib/_dialog.scss
+++ b/dialog/lib/_dialog.scss
@@ -241,7 +241,7 @@
     transition-timing-function: var(--_closing-transition-easing);
   }
 
-  :host([trasition][closing]) .container > * {
+  :host([transition][closing]) .container > * {
     transform: none;
     opacity: 0;
   }
--- a/dialog/lib/dialog.ts
+++ b/dialog/lib/dialog.ts
@@ -393,7 +393,8 @@ export class Dialog extends LitElement {
     let promise = this.updateComplete;
     if (duration > 0) {
       promise = new Promise((r) => {
-        setTimeout(r, duration);
+        // setTimeout(r, duration);
+        this.containerElement.ontransitionend = () => r(null);
       });
     }
     await promise;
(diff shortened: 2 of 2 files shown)


vdegenne:
@e111077 @asyncLiz I spent quite some time on this one because I didn't want to just throw in some changes until it worked, I analyzed the existing state of this element and came up with that solution.
I think it's an important update because it makes the dialog visually more *stable*.
I would really appreciate if you could review this PR as soon as you guys can.

For context, this fixes the issue I was talking some days ago on Discord : https://discord.com/channels/1012791295170859069/1113583575808344106/1113583579763572886

Thanks

dialog/lib/dialog.ts
```diff
@@ -389,7 +393,8 @@ export class Dialog extends LitElement {
     let promise = this.updateComplete;
     if (duration > 0) {
       promise = new Promise((r) => {
-        setTimeout(r, duration);
+        // setTimeout(r, duration);
+        this.containerElement.ontransitionend = () => r(null);
```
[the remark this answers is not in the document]
asyncLiz: Can we actually revert this change and just use the promise awaiting for now?

Using `on<event>` properties is generally frowned upon, and I'd like to avoid any unknown side effects from changing this since it doesn't impact animation.

dialog/lib/dialog.ts
```diff
@@ -349,7 +349,11 @@ export class Dialog extends LitElement {
     }
     // Avoids dispatching initial state.
     const shouldDispatchAction = changed.get('open') !== undefined;
-    this.performTransition(shouldDispatchAction);
+
+    // Avoids performing unnecessary transition
+    if (changed.get('open') !== undefined && changed.get('open') !== this.open) {
```
asyncLiz, line 354: `shouldDispatchAction` is the same value as the first condition in this if block. Is that intentional? Would be the same as

```ts
if (changed.get('open') !== undefined && changed.get('open') !== this.open) {
  this.performTransition(true);
}
```

Are there other places that `performTransition` is called with a false value?

asyncLiz (changes requested):

dialog/lib/_dialog.scss
```diff
@@ -192,7 +194,7 @@
   // Transitions for open/closed states
   .container {
     will-change: transform, opacity;
-    transition-property: transform;
+    transition-property: transform, opacity;
```
asyncLiz, line 197: I checked the spec, and opacity's timing function should be `linear`. Can you add that below?

```scss
transition-timing-function: var(--_closing-transition-easing), linear;
```

asyncLiz: Hm, it also has a kind of annoying content fade out before container fade out situation...

Do you mind adding a comment above `transition-property`?

```scss
// TODO: make opacity fade out content and container separately
```

asyncLiz:
Mind rebasing for the conflicting file? :)

vdegenne:
@asyncLiz can I resolve conflicts from Github interface and merge instead?

asyncLiz:
> @asyncLiz can I resolve conflicts from Github interface and merge instead?

Maaaybe? One way to find out lol

vdegenne:
@asyncLiz That did it!

dialog/lib/dialog.ts
```diff
@@ -383,17 +389,17 @@ export class Dialog extends LitElement {
 
   private dialogClosedResolver?: () => void;
 
-  private async performTransition(shouldDispatchAction: boolean) {
+  private async performTransition(/*shouldDispatchAction: boolean*/) {
```
asyncLiz, line 392: @sorvell do you have more context on when we should or should not dispatch the action event on a transition?

asyncLiz:
Obsolete now with dialog's refactor in https://github.com/material-components/material-web/pull/4677

PR #4503, opened 2023-06-27
Update commitlint.yml

M1c4h20:


asyncLiz:
We use commitlint to ensure proper commit messages, which are used for release notes :)

Aamir333:
Thanks boss

PR #4505, opened 2023-06-28
chore(chip): reflect filter chip selected value

vdegenne:
In a chip set context, there are currently no way to query selected chips from the outside  
(e.g. `queryAll('md-filter-chip[selected]')`).

This PR solves that problem.

### DIFF
--- a/chips/lib/filter-chip.ts
+++ b/chips/lib/filter-chip.ts
@@ -20,7 +20,7 @@ import {renderRemoveButton} from './trailing-icons.js';
 export class FilterChip extends MultiActionChip {
   @property({type: Boolean}) elevated = false;
   @property({type: Boolean}) removable = false;
-  @property({type: Boolean}) selected = false;
+  @property({type: Boolean, reflect: true}) selected = false;
 
   protected get primaryId() {
     return 'option';
--- a/chips/lib/input-chip.ts
+++ b/chips/lib/input-chip.ts
@@ -20,7 +20,7 @@ export class InputChip extends MultiActionChip {
   @property() href = '';
   @property() target: '_blank'|'_parent'|'_self'|'_top'|'' = '';
   @property({type: Boolean, attribute: 'remove-only'}) removeOnly = false;
-  @property({type: Boolean, refect: true}) selected = false;
+  @property({type: Boolean, reflect: true}) selected = false;
 
   protected get primaryId() {
     if (this.href) {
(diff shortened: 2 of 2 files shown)


asyncLiz:
I'm curious, what's the context where you need to querySelector the chips instead of using the chip-set's `chips` property?

```ts
const chipSet = document.querySelector('md-chip-set');
const selectedChips = chipSet.chips.filter(chip => (chip as MdFilterChip).selected);
```

vdegenne:
> I'm curious, what's the context where you need to querySelector the chips instead of using the chip-set's `chips` property?
> 
> ```ts
> const chipSet = document.querySelector('md-chip-set');
> const selectedChips = chipSet.chips.filter(chip => (chip as MdFilterChip).selected);
> ```

Sorry I didn't know about that one.  
But don't you think it's an interesting improvement? It supports declarative querying.

asyncLiz:
Yes, though when will a client need that? The only place I can think of is CSS, but we have dedicated "selected" tokens for different selection styling.

I'm not strongly opposed to it right now, but I know we've had to adjust reflecting the attributes of some components to mimic native behavior. Notably, `<select>`'s `<option>` does not reflect its selected property to an attribute.

> Side note: I want to adopt the [element internals states api](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals/states) for state querying as a future feature.
>
> ```ts
> const selectedChips = document.querySelectorAll('md-filter-chip:--selected');
> const removableChips = document.querySelectorAll('md-filter-chip:--removable');
> ```

vdegenne:
I don't really see a case where one would want to style a chip using CSS, maybe for changing the opacity of unselected chips??  
Anyway I just thought it was more intuitive, the solution you provided doesn't bother me that much, it's just the query selector solution was the first I tried without jumping into the doc or into the source code.  
We can use `selected` attribute to initially select chips, I just thought clients would naturally think it's reflected?

Another thing, let's imagine there are many chipsets for different subjects, you can query all selected chips using the selector version, whereas the JS solution would require to concat all filtered chips from all queried chipset...

chips/lib/filter-chip.ts
```diff
@@ -20,7 +20,7 @@ import {renderRemoveButton} from './trailing-icons.js';
 export class FilterChip extends MultiActionChip {
   @property({type: Boolean}) elevated = false;
   @property({type: Boolean}) removable = false;
-  @property({type: Boolean}) selected = false;
+  @property({type: Boolean, reflect: true}) selected = false;
```
asyncLiz, line 23: Can you add this to input chip's `selected` attribute as well?

asyncLiz (changes requested):
I think this makes sense to go with for now. Ultimately "selected" is a state that can be changed by the chip itself, which is a good candidacy reason for reflecting the attribute.

chips/lib/filter-chip.ts
vdegenne: Ok nice, I just updated.

PR #4506, opened 2023-06-28
fix: wireit sass watch memory leak

vdegenne:
This will avoid wireit looping through the `@material/web` symbolic link in watch mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,9 +98,10 @@
       ]
     },
     "build:sass": {
-      "command": "sass --style=compressed --load-path=node_modules --load-path=node_modules/sass-true/sass $(ls -d */ | grep -v node_modules)",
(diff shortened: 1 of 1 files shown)


e111077:
OMG wow! great find!

e111077 (approved):

vdegenne:
@e111077 wait!

e111077:
if there's a change you need to be made, you'll need to start another PR since the merge process is largely automated

vdegenne:
@e111077  No I just wanted to say this solution works in the case where `catalog` is removed from `workspaces` and has its own `node_modules` directory.
Accepting this change in the current state of this repo will actually have no effect, it's still good for prevention though.  

IMO yesterday @augustjk's PR ( https://github.com/material-components/material-web/pull/4501) is a better solution but one needs to be aware it only works if `package-lock.json` is removed before `npm i`.

PR #4509, opened 2023-06-29
fix: catalog dev watch memory leak

vdegenne:
Following https://github.com/material-components/material-web/pull/4506 which didn't really solve the memory leak, just slowed it down.

This PR really solves the memory leak problem which originated from the `copy-stories` wireit task in the `catalog` package.  
`../*/demo` value in `files` field was expanding to `../node_modules/...` paths, causing the loop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "!catalog"
   ],
   "workspaces": [
+    ".",
     "catalog"
   ],
   "dependencies": {
--- a/catalog/package.json
+++ b/catalog/package.json
@@ -146,7 +146,8 @@
       "command": "node scripts/copy-stories.mjs",
       "files": [
         "scripts/copy-stories.mjs",
-        "../*/demo"
+        "../*/demo",
+        "!../node_modules"
       ],
       "output": [
         "stories/*/**/*",


vdegenne:
@augustjk Requesting a review from you since we've discussed about this issue together.

package.json
```diff
@@ -100,8 +101,7 @@
     "build:sass": {
       "command": "sass --style=compressed --load-path=node_modules --load-path=node_modules/sass-true/sass $(ls -d */ | grep -vE 'node_modules|catalog')",
       "files": [
-        "**/*.scss",
-        "!catalog"
```
e111077, line 104: Why is this one necessary to remove?

e111077 (changes requested):

package.json
vdegenne: Because the catalog doesn't contain `scss` files **and** the root is back to `workspaces` therefore the `@material/web` package is symlinked in root `node_modules` which is a directory ignored by wireit by default.  

I could add it back if you want, but it works fine without.

e111077: I think you should add it back in case we do add scss as it would likely be a sub-command inside the catalog dir

vdegenne: Ok it makes sense. I reverted.

e111077 (approved):

e111077:
Cool, sent for internal review. Unfortunately I can't approve it because I imported it, and it's currently outside of regular working hours, so someone should approve it tomorrow and the bot should merge.

PR #4548, opened 2023-07-12
chore(checkbox): add checked event

vdegenne:
Would you all be interested in providing a `checked` event for the `MdCheckbox`?

### DIFF
--- a/checkbox/lib/checkbox.ts
+++ b/checkbox/lib/checkbox.ts
@@ -14,6 +14,19 @@ import {classMap} from 'lit/directives/class-map.js';
 import {ARIAMixinStrict} from '../../internal/aria/aria.js';
 import {requestUpdateOnAriaChange} from '../../internal/aria/delegate.js';
 import {dispatchActivationClick, isActivationClick, redispatchEvent} from '../../internal/controller/events.js';
+import type {MdCheckbox} from '../checkbox.js';
+
+export class CheckboxCheckedEvent extends Event {
+  constructor (public checkbox: MdCheckbox) {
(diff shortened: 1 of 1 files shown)


checkbox/lib/checkbox.ts
```diff
@@ -110,6 +110,11 @@ export class Checkbox extends LitElement {
       this.prevDisabled = changed.get('disabled') ?? this.disabled;
       this.prevIndeterminate =
           changed.get('indeterminate') ?? this.indeterminate;
+      
+      this.dispatchEvent(new CustomEvent('checked', {
```
christophe-g, line 114: Would the `input` event not be enough in this case?
```js
checked.addEventListener("input", updateValue);

function updateValue(e) {
  console(e.target.checked);
}```

vdegenne: @christophe-g The problem is the same with `onChange`, if you change `checkbox.checked` manually it doesn't subsequently call `change` or `input`.
I don't think this is an issue because that's the native behavior with html native inputs, that's why I thought having a dedicated event (with detail value) for that would be handy.

christophe-g: @vdegenne  - ah d'accord

I think the tendency here is to follow native behaviors - hence only dispatch events when user interact with the element (e.g. click or keyboard input).

asyncLiz:
Unfortunately no, for the most part we should only be dispatching events due to user interactions.

I also think we got a bit burnt by dispatching programmatic events in previous implementations.

PR #4560, opened 2023-07-17
set max height/width for iconbutton

### LINKED ISSUE #2457: mwc-button rendering different on Safari to Chrome
**Describe the bug**
In Safari the button has additional 2px margin/padding to both the left and right.

In additon, the button label text does not appear to be correctly vertically aligned with the icon in **IOS Mobile Safari only**.

**To Reproduce**
Either fully left or right align button and you'll see a 2px gap in Safari. However, in Chrome, it there is no gap.

In **IOS Mobile Safari only** you will see trailing icon and label text are not vertically aligned correctly.

**Expected behavior**
No left or right margin/padding.

Icon and label text vertically aligned correctly in IOS Mobile Safari.

**Screenshots**
Chrome (renders correct)
image

Safari (renders incorrect, 2px gap to right of button)
image

IOS Mobile Safari only ('Templates' label text not vertically aligned correctly)

Code uses 'menu' Material Icon
```
<mwc-button label="Templates" title="Template">
<svg slot="trailingIcon" xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px">
<path d="M0 0h24v24H0V0z" fill="none" />
<path d="M3 18h18v-2H3v2zm0-5h18v-2H3v2zm0-7v2h18V6H3z" />
</svg>
</mwc-button>
```

image

Safari Version 14.1
IOS Safari, version 14.5.1
Version 90.0.4430.212

### LINKED ISSUE #4497: [md-icon-button] Incorrect spacing on iOS Safari
Is it possible to add support for `spacing-leading` and `spacing-trailing` tokens? 
This is because browser's default stylesheet is causing varied behaviour rendering the `md-*-icon-button`.

**Screenshots**
Rendered on Chrome 114
image 1

Rendered on iOS 16.4 Safari
image

### LINKED ISSUE #4559: md-standard-icon-button getting squooshed
image

VandeurenGlenn:
fixes #4559
fixes #4497 
fixes #2457


iconbutton/lib/_standard-icon-button.scss
```diff
@@ -35,8 +35,10 @@
       --_#{$token}: var(--md-icon-button-#{$token}, #{$value});
     }
 
-    height: var(--_state-layer-size);
-    width: var(--_state-layer-size);
+    max-height: var(--_state-layer-size);
+    max-width: var(--_state-layer-size);
+    width: 100%;
```
asyncLiz, line 40: Icon buttons should not expand to fill the width/height of their container, so I think using width/height for the state layer size is correct.

Can you give some examples where max-* is needed? #4559 looks like it might be flexbox-related

VandeurenGlenn: @asyncLiz 
It's a Mozilla and Apple thing, if you don't set min or max then they will ignore height or width.

VandeurenGlenn: Iow setting min/max and height/width to 100% does what we expect width/height does on chrome

VandeurenGlenn: https://lit.dev/playground/#gist=7c7c9aa6d2a05b5f273d60df84c8a91e

VandeurenGlenn: https://lit.dev/playground/#gist=7c7c9aa6d2a05b5f273d60df84c8a91e

asyncLiz: This is what I'd expect with a `width: 100%` item in a flex container with other siblings.

You can fix this by setting `flex: 1` on the filler instead of `width: 100%`. That tells flex box to fill the remaining space, while not trying to squish other siblings to try and make the filler truly "100%" the width of the container.

Alternatively, you can set `flex-shrink: 0` on specific items to tell the flex box not to squish them.

VandeurenGlenn: @asyncLiz OMG, my bad! thanks for checking it out!,`flex-shrink` cool learned something new!

PR #4617, opened 2023-07-27
build(catalog): use firebase for hosting and remove url prefixes

e111077:
creates a firebase action that deploys on main and creates a github actions bot that deploys previews on PR.


AndrewJakubowicz:
Nice! I assume this is working because the preview URL looks great!

PR #4640, opened 2023-07-30
fix(icon)!: set height/width on host

### LINKED ISSUE #4639: No height/width set on md-icon
https://github.com/material-components/material-web/blob/2568d4fbac69c8b5e9a5d5198c6525224ebf778d/icon/internal/_icon.scss#L29

Setting the height and with to svg but not the element itself disables setting the desired width/height on the slotted svg element

VandeurenGlenn:
This sets width/height to host and making slotted content inherit those.

fixes #4639


asyncLiz:
Fixed now!

PR #4649, opened 2023-08-02
chore(textfield): remove unnecessary redispatchEvent

vdegenne:


textfield/harness.ts
```diff
@@ -8,6 +8,11 @@ import {Harness} from '../testing/harness.js';
 
 import {TextField} from './internal/text-field.js';
 
+export const inputEventInitBase: InputEventInit = {
```
asyncLiz, line 11: Can you inline this where it's used rather than creating a shared variable? It'll make the tests a bit more damp and help with scanning code to see what it's doing

asyncLiz (changes requested):

textfield/harness.ts
vdegenne: Sure!

asyncLiz (approved):

PR #4671, opened 2023-08-07
fix(list): list item body overflow issue

davidurco:
Hey there!

Excited to be contributing to this repository for the first time!

So, in this pull request, I tackled an issue with overflowing content in the list-item's body. The fix was pretty straightforward - I set `max-width: 100%` to the list-item's body element. To make things even better, I also got rid of the `display: flex` from `label-text`, which wasn't doing much except making truncation more complicated. I mean, it's just text inside label-text, so no need for flex, right?

Here's how it looks like:

Screenshot showing the fixed list-item's body content

Many thanks for checking this out! Looking forward to your review!

Cheers,
David

### DIFF
--- a/list/internal/listitem/_list-item.scss
+++ b/list/internal/listitem/_list-item.scss
@@ -134,7 +134,7 @@
     display: inline-flex;
     justify-content: center;
     flex-direction: column;
-    max-width: 100%;
+    min-width: 0;
     box-sizing: border-box;
     flex: 1 0 0;
     padding-inline-start: var(--_list-item-leading-space);
--- a/list/internal/listitem/_list-item.scss
+++ b/list/internal/listitem/_list-item.scss
@@ -134,6 +134,7 @@
     display: inline-flex;
     justify-content: center;
     flex-direction: column;
+    max-width: 100%;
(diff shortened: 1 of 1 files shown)


list/internal/listitem/_list-item.scss
```diff
@@ -134,6 +134,7 @@
     display: inline-flex;
     justify-content: center;
     flex-direction: column;
+    max-width: 100%;
```
asyncLiz, line 137: `max-width: 100%` will make the text overflow if there's also start or end content. This is because the `.body` is trying to take up the whole width of the flex container.

Change this to `min-width: 0` and watch the magic happen.

See https://stackoverflow.com/a/66689926 for a more in-depth explanation as to why this works.

asyncLiz (changes requested):

list/internal/listitem/_list-item.scss
davidurco: @asyncLiz Oh, of course. I just updated the PR. Thanks for the review!

asyncLiz (approved):
Thanks for the fix! Sorry for the delay in reviewing

PR #4694, opened 2023-08-11
fix(list-item): rename selector

MrHBS:
I renamed content-wrapper to container since the latter made more sense to me and also matches the naming used in TextField. You may close this PR if you disagree 😅


asyncLiz:
Thank you for the suggestion! I think it's ok as-is for now though

PR #4755, opened 2023-08-20
fix(dialog): open on default

VandeurenGlenn:
ensure the dialog has been updated before showing


asyncLiz:
Fixed in https://github.com/material-components/material-web/pull/4759

PR #4768, opened 2023-08-23
fix(icon): provide font-variation-settings token

vdegenne:


asyncLiz:
This is WAI, for changing font-variation-settings we recommend using an initializer or shared CSS

```ts
const sharedStyles = css`
  md-icon.filled {
    font-variation-settings: 'FILL' 1;
  }
`;

class MyComponent extends LitElement {
  override static styles = [sharedStyles, css`...`];

  override render() {
    return html`
      <md-icon class="filled">event</md-icon>
      ...
    `;
  }
}
```

```ts
import {MdIcon} from '@material/web/icon/icon.js';

MdIcon.addInitializer(instance => {
  instance.style.fontVariationSettings = "'FILL' 1";
});
```

PR #4770, opened 2023-08-23
fix(dialog): prevent close when click released in scrim

vdegenne:
This is an attempt to avoid the dialog to close when the mouse click was initiated inside the content but released from the scrim.

### DIFF
--- a/dialog/internal/dialog.ts
+++ b/dialog/internal/dialog.ts
@@ -238,7 +238,7 @@ export class Dialog extends LitElement {
         .returnValue=${this.returnValue || nothing}
       >
         <div class="container"
-          @click=${this.handleContentClick}
+          @mousedown=${this.handleContentMouseDown}
         >
           <div class="headline">
             <div class="icon" aria-hidden="true">
@@ -278,8 +278,15 @@ export class Dialog extends LitElement {
     this.intersectionObserver.observe(this.bottomAnchor!);
   }
 
-  private handleDialogClick() {
+  private handleDialogClick(event: MouseEvent) {
+
     if (this.nextClickIsFromContent) {
+      if (event.target === this.dialog) {
+        // Click was pressed inside the content
+        // but released outside content.
+        // Prevent closing.
+        event.preventDefault();
+      }
       // Avoid doing a layout calculation below if we know the click came from
       // content.
       this.nextClickIsFromContent = false;
@@ -297,7 +304,7 @@ export class Dialog extends LitElement {
     this.close();
   }
 
-  private handleContentClick() {
+  private handleContentMouseDown() {
     this.nextClickIsFromContent = true;
   }
 


dialog/internal/dialog.ts
```diff
@@ -238,7 +238,7 @@ export class Dialog extends LitElement {
         .returnValue=${this.returnValue || nothing}
       >
         <div class="container"
-          @click=${this.handleContentClick}
+          @mousedown=${this.handleContentMouseDown}
```
asyncLiz, line 241: Can we swap this to `pointerdown`?

asyncLiz (changes requested):

dialog/internal/dialog.ts
vdegenne: And preserving the existing `click` event you mean?

vdegenne: I swapped to pointerdown and renamed the flag so it reads better.

asyncLiz (approved):

asyncLiz:
Update on this, there's some screenshot test failures that need more investigating, so apologies for the delay

vdegenne:
> Update on this, there's some screenshot test failures that need more investigating, so apologies for the delay

Can you give me more details? the tests run fine on my end.

asyncLiz:
> > Update on this, there's some screenshot test failures that need more investigating, so apologies for the delay
> 
> Can you give me more details? the tests run fine on my end.

It's internal tests, so they don't run on github 😔

vdegenne:
I noticed quite a few quirks on the `select` element (not just in a dialog), maybe it's a good opportunity to polish this element instead of trying to fix one problem related to this specific case.

e111077:
the failing screenshot test seems to be weird. I can't nail down the exact issue but looks flaky when attempting to reproduce the issue with a debugger in a live environment in the test. Otherwise, normal interaction looks good. Gonna have to spend a bit of time to figure out what's going on here

dialog/internal/dialog.ts
VandeurenGlenn: Doesn't click return the target where mouse up happens?

austinw-fineart:
I've got a failing case that's easy to reproduce:
1. Put a checkbox in the dialog
2. Toggle the checkbox with the Spacebar
3. Watch it close the dialog

Not every click event has a preceding pointerdown event so the next click will fail the check. You should probably just handle both events.

vdegenne:
@austinw-fineart This looks like a different issue to me, plus I couldn't reproduce it.
If you have a problem could you open another issue and join a playground to it?
You can use this template: https://lit.dev/playground/#gist=db372518cfa9cd6097b7842e6da5f740
Thanks

austinw-fineart:
@vdegenne I'm afraid I do not know how to import this into playground. I ran the test case on this pull request.

PR #4835, opened 2023-09-02
fix(badge): center align value text

vdegenne:
I noticed the text in the badge was not properly centered. This is the first solution that came to my mind, any thought on that?

image


asyncLiz (approved):
Sounds good for now since this is still in progress

PR #4838, opened 2023-09-04
chore(text-field): add autocomplete attribute

### LINKED ISSUE #82: Broken textfield autocomplete
The autosuggest fails until input tag is inside form tag.
https://github.com/material-components/material-components-web/issues/3174

But It doesn't work for form level autosuggest. https://github.com/w3c/webcomponents/issues/572

### LINKED ISSUE #4837: md-text-field ignoring autocomplete attribute
It seems that `autocomplete="off"` on textfields is not working properly, and do not seem to turn off the autocompletion.  
I am not sure how to reproduce that, but setting `autocomplete="off"` manually on the internal `input` element gets rid of the problem.  
Which means we should probably find a way to explicitly relay the value from the outside.

vdegenne:
Fixes #4837
Fixes #82

### DIFF
--- a/textfield/internal/text-field.ts
+++ b/textfield/internal/text-field.ts
@@ -36,23 +36,6 @@ export type InvalidTextFieldType =
     'button'|'checkbox'|'hidden'|'image'|'radio'|'range'|'reset'|'submit';
 
 /**
- * autocomplete types for the text field.
- */
-export type TextFieldAutoCompleteType =
-    'on'|'off'|'additional-name'|'address-level1'|'address-level2'|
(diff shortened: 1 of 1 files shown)


asyncLiz (approved):
🤌

asyncLiz:
Yea it's mad at us too. Try swapping it to this:

```
// Lit analyzer has too strict of types for input autocomplete
// tslint:disable-next-line:no-any
const autocomplete = this.autocomplete as any;

<input autocomplete=${autocomplete || nothing}
```

We also need to add the attribute to the `<textarea>`

asyncLiz (approved):

Issue #4879, opened 2023-09-08
add "all.js" as main/module in package.json so it could be used by a designer


jogibear9988:
Hello,

I'm workin on a Designer Framework, wich can automaticly load NPM modules with webcomponents. (see https://node-projects.github.io/web-component-designer-demo/index.html)

I got now your webcomponents workin, but only after adding a hack (https://github.com/node-projects/web-component-designer/blob/master/src/elements/helper/NpmPackageHacks.json#L9), so that the all.js is loaded when importing the npm (cause all.js is not in the package.json as main or module export).

It would also be possible if you add a custom-elments.json wich describes all your elements, but as long this is not htere yet, a main/module export will work.

jogibear9988:
workin with the hack:

<img width="566" alt="image" src="https://github.com/material-components/material-web/assets/364896/2e3cfa6f-5a38-40f7-b6e6-5cfb39c77098">

jogibear9988:
https://github.com/material-components/material-web/pull/4880

asyncLiz:
Unfortunately there's two good reasons why we didn't do this:

1. We don't want to encourage loading all components in production (only for dev projects).
2. We want to reserve the main entrypoint for future APIs.

We plan on providing side-effect-free imports, which _will_ be the main entrypoint. For example:

```js
import {Checkbox, styledCheckbox} from '@material/web';

customElements.define('my-checkbox', styledCheckbox(css`...`));

const element = document.querySelector<Checkbox>('my-checkbox');
```

> It would also be possible if you add a custom-elments.json

I think this would be the right approach. We have https://github.com/material-components/material-web/issues/545 to track that.

PR #4880, opened 2023-09-08
Add main export

jogibear9988:


jogibear9988:
fixes #4879

asyncLiz:
See https://github.com/material-components/material-web/issues/4879#issuecomment-1712285128 for why we want to reserve the main entrypoint

PR #4889, opened 2023-09-10
fix (select): close select on menu close

43081j:
There are some situations where the menu can close when the outer select is still open and unaware.

For example, clicking the field will open the menu, then clicking it again will close the menu **but not the select**.

This is because the menu observes outside clicks in order to close itself. We don't currently observe those closes to update our own state inside `md-select`, so the UI still shows as if it is open but without a menu.

The fix here is to always reflect the state of the menu on `closed`.

Example of this bug:

- Go [here](https://lit.dev/playground/#project=W3sibmFtZSI6ImJsb29wLnRzIiwiY29udGVudCI6ImltcG9ydCAnQG1hdGVyaWFsL3dlYi9zZWxlY3Qvc2VsZWN0LW9wdGlvbi5qcyc7XG5pbXBvcnQgJ0BtYXRlcmlhbC93ZWIvc2VsZWN0L291dGxpbmVkLXNlbGVjdC5qcyc7In0seyJuYW1lIjoiaW5kZXguaHRtbCIsImNvbnRlbnQiOiI8IURPQ1RZUEUgaHRtbD5cbjxoZWFkPlxuICA8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCIuL2Jsb29wLmpzXCI-PC9zY3JpcHQ-XG48L2hlYWQ-XG48Ym9keT5cbjxtZC1vdXRsaW5lZC1zZWxlY3Q-XG4gICAgICA8bWQtc2VsZWN0LW9wdGlvbiB2YWx1ZT1cIm9uZVwiIGhlYWRsaW5lPVwiT25lXCI-PC9tZC1zZWxlY3Qtb3B0aW9uPlxuICAgICAgPG1kLXNlbGVjdC1vcHRpb24gdmFsdWU9XCJ0d29cIiBoZWFkbGluZT1cIlR3b1wiPjwvbWQtc2VsZWN0LW9wdGlvbj5cbjwvbWQtb3V0bGluZWQtc2VsZWN0PlxuPC9ib2R5PlxuIn0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjIuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMS4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4yLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9XQ)
- Click the select box
- Click it again
- Observe that it thinks it is still open (e.g. the arrow is pointing the wrong way)


select/harness.ts
```diff
@@ -48,6 +51,18 @@ export class SelectHarness extends Harness<Select> {
     field.click();
   }
 
+  async clickAndWaitForMenu() {
```
asyncLiz, line 54: Could we override `click()` and wait for the menu's events there? I'm not sure when you'd want to click the element and not wait for all the events to settle.

select/internal/select.ts
```diff
@@ -642,6 +642,14 @@ export abstract class Select extends selectBaseClass {
     return hasSelectedOptionChanged;
   }
 
+  /**
+   * Closes the select if the inner menu closed
+   */
+  private handleClosed(e: Event) {
```
asyncLiz, line 648: I think Elliott [added this](https://github.com/material-components/material-web/blob/ec0a8ebf6ec973282b9a382a1f2fa347b5c82975/select/internal/select.ts#L683-L686) in https://github.com/material-components/material-web/commit/78e7c1742f04437eee3f804cc73de9d3ffaaf1bb right?

asyncLiz (changes requested):

select/internal/select.ts
43081j: aha so he did, 2 days after i opened this. didn't see that, my mistake.

this is just a tests PR in that case, he already pushed the same fix.

PR #4891, opened 2023-09-10
test(button): add tests for button internals

43081j:
This just adds some tests for the underlying button implementation

can you let me know if this is how you've been testing internals? so i can keep it consistent with the rest of the repo

basically i used text-button for sake of getting a button but tested the internal behaviour all of them have


button/internal/button_test.ts
```diff
@@ -0,0 +1,215 @@
...
+  beforeEach(() => {
+    root = document.createElement('div');
+    document.body.appendChild(root);
+  });
+
+  afterEach(() => {
+    root?.remove();
+  });
+
+  it('renders inner button', async () => {
```
asyncLiz, line 23: These tests seem really granular and test the implementation rather than the functionality.

What sort of value are we hoping to get out of these tests?

43081j: we don't seem to have any button tests yet.

i agree these are fairly granular as they are basically asserting we rendered what we expected to render.

you could try test from the outside instead, e.g. test that a button with a `href` is clickable as a link. though that seems we will just end up back here with less assertions (but the same remaining assertions).

one way or another, in that example, you have to test that the `href` actually resulted in something. so you have to assert we opened the link after clicking the button, or that the link existed in DOM.

PR #4904, opened 2023-09-12
fix(tabs): apply shape to tab elevation

bbuhler:
The elevation of a tab is displayed in square shape:
image

This fix is applying the shape to `md-elevation` element as well:
image


asyncLiz (approved):
Thanks for this fix!

PR #4922, opened 2023-09-14
Corrected button trailing-icon

datvm:
On https://material-web.dev/components/button/, both the icons are at the front. The correct attribute is `trailing-icon`. Source code:

https://github.com/material-components/material-web/blob/573caaee1b7fe64b57b84e2b287b6c5c74666de1/button/internal/button.ts#L59


datvm:
To maintainers: I already signed the CLA but clicking Rescan results in 400 Bad Request error.

PR #5114, opened 2023-10-24
fix(all): add chip-set to all.js file

vdegenne:
This PR will add `chip-set` element to the list in `all.js`.


asyncLiz (changes requested):
LGTM, can you add it to `common.js` as well?

asyncLiz (approved):

AndrewJakubowicz:
Thank you! Quick nit: internally lint is failing because those imports need to be in alphabetical order.

So `chip-set` needs to be moved down one line between `assist-chip` and `filter-chip`.

vdegenne:
> Thank you! Quick nit: internally lint is failing because those imports need to be in alphabetical order.
> 
> So `chip-set` needs to be moved down one line between `assist-chip` and `filter-chip`.

I knew there were something with these lint comments 😹
I've updated.

PR #5212, opened 2023-11-22
chore(segmented-button): make tap highlight transparent

vdegenne:


asyncLiz (approved):
Gonna update the title to `chore` since segmented button is still in labs. Chores won't show up on the changelog. That way folks don't think segmented button has been shipped

vdegenne:
> Gonna update the title to `chore` since segmented button is still in labs. Chores won't show up on the changelog. That way folks don't think segmented button has been shipped

Ah great! I didn't know this keyword had a post influence, learning everyday!

PR #5229, opened 2023-11-29
fix(menu): prevent native popover text color from overriding menu surface text color

aarondrabeck:
fixes: https://github.com/material-components/material-web/issues/5226


e111077 (approved):

e111077 (approved):
running secondary tests

PR #5263, opened 2023-12-06
feat(chip): trailing remove icon can now be customized

vdegenne:
I made a preliminary commit just to show the course it's taking.

A `slot` was inserted directly into the `renderRemoveButton` method so everything works as intended with the keydown events and focus logic.

Another advantage is that `md-input-chip` will also inherit this change so users can use a different remove trailing icon if they want.

### DIFF
--- a/chips/internal/_trailing-icon.scss
+++ b/chips/internal/_trailing-icon.scss
@@ -11,7 +11,8 @@
   .trailing.action {
     align-items: center;
     justify-content: center;
-    padding: 0 8px;
+    padding-inline-start: var(--_icon-label-space);
+    padding-inline-end: var(--_with-trailing-icon-trailing-space);
   }
 
   .trailing.action :is(md-ripple, md-focus-ring) {
--- a/chips/internal/_trailing-icon.scss
+++ b/chips/internal/_trailing-icon.scss
@@ -35,7 +35,7 @@
     width: var(--_icon-size);
   }
 
-  [name="trailing-icon"]::slotted(md-icon) {
+  [name="remove-trailing-icon"]::slotted(md-icon) {
      --md-icon-size: var(--_icon-size);
      color: inherit;
   }
--- a/chips/internal/_trailing-icon.scss
+++ b/chips/internal/_trailing-icon.scss
@@ -35,11 +35,6 @@
     width: var(--_icon-size);
   }
 
-  [name="remove-trailing-icon"]::slotted(md-icon) {
-     --md-icon-size: var(--_icon-size);
-     color: inherit;
-  }
-
   :where(:hover) .trailing.icon {
     color: var(--_hover-trailing-icon-color);
   }
--- a/chips/internal/_trailing-icon.scss
+++ b/chips/internal/_trailing-icon.scss
@@ -28,7 +28,7 @@
     padding-inline-end: 0;
   }
 
-  .trailing.icon, [name="trailing-icon"]::slotted(*) {
+  .trailing.icon {
     color: var(--_trailing-icon-color);
     height: var(--_icon-size);
     width: var(--_icon-size);
@@ -36,6 +36,7 @@
 
   [name="trailing-icon"]::slotted(md-icon) {
      --md-icon-size: var(--_icon-size);
+     color: inherit;
   }
 
   :where(:hover) .trailing.icon {
--- a/chips/internal/_trailing-icon.scss
+++ b/chips/internal/_trailing-icon.scss
@@ -28,10 +28,14 @@
     padding-inline-end: 0;
   }
 
-  .trailing.icon {
+  .trailing.icon, [name="trailing-icon"]::slotted(*) {
     color: var(--_trailing-icon-color);
     height: var(--_icon-size);
     width: var(--_icon-size);
+  }
+
+  [name="trailing-icon"]::slotted(md-icon) {
+     --md-icon-size: var(--_icon-size);
   }
 
   :where(:hover) .trailing.icon {
--- a/chips/internal/trailing-icons.ts
+++ b/chips/internal/trailing-icons.ts
@@ -34,10 +34,12 @@ export function renderRemoveButton({
       @focus=${focusListener}>
       <md-focus-ring part="trailing-focus-ring"></md-focus-ring>
       <md-ripple ?disabled=${disabled}></md-ripple>
-      <svg class="trailing icon" viewBox="0 96 960 960" aria-hidden="true">
-        <path
-          d="m249 849-42-42 231-231-231-231 42-42 231 231 231-231 42 42-231 231 231 231-42 42-231-231-231 231Z" />
-      </svg>
+      <slot name="trailing-icon">
+        <svg class="trailing icon" viewBox="0 96 960 960" aria-hidden="true">
+          <path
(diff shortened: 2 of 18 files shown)


vdegenne:
I think it's going well,
I added a getter to determine if user inserted a customized trailing icon, I then use this flag to block the default `remove` event, *rather than blocking we should probably dispatch another custom event instead, `trailing-icon-click` for instance*

I will wait for your feedback, now.

asyncLiz:
Also let's check in on the motivation for this change. Do we want to replace the remove button's icon with something else, like a trash can? Or do we want to replace remove functionality with something new, such as a dropdown icon?

vdegenne:
@asyncLiz It seems that the more general the better in this case, I see some cases where one developer would want to do something more than just removing or showing a dropdown icon (for instance showing an "info" icon that opens a small dialog to give more info about a special type of filter)

asyncLiz:
I agree! Let's tackle that separately then, and scope this to just customizing the remove button's icon :)

I'll update the pr title to reflect

asyncLiz (changes requested):
omg I forgot to publish my comments... so sorry!

vdegenne:
@asyncLiz Ok so if I understand you want to give the ability to change remove icon and later add another slot so users can insert anything at the end?
That means this code
```html
<md-filter-chip
  removable
>
  <md-icon-button slot="trailing-icon">...</md-icon-button>
</md-filter-chip>
```
will end up showing two different trailing elements, is that ok?

asyncLiz:
> will end up showing two different trailing elements

I'm thinking that `removable` will not have an effect if a user provides a new trailing action.

```html
<!-- Only the new icon button shows, not a second one. `removable` has no effect -->
<md-filter-chip removable>
  <md-icon-button slot="trailing-icon">...</md-icon-button>
</md-filter-chip>

<!-- Remove button shows with different icon -->
<md-filter-chip removable>
  <md-icon slot="remove-trailing-icon">trash</md-icon>
</md-filter-chip>

<!-- Only the new icon button shows, custom remove trailing icon does not. -->
<md-filter-chip removable>
  <md-icon slot="remove-trailing-icon">trash</md-icon>
  <md-icon-button slot="trailing-icon">...</md-icon-button>
</md-filter-chip>
```

asyncLiz:
> Just to make sure we are on the same page, the modifications I provide allow this writing:
> 
> ```
> 
> <md-filter-chip
>   has-trailing
>   @remove=${/* executed when remove icon button clicked */}
> ></md-filter-chip>
> 
> 
> <md-filter-chip
>   has-trailing
>   @remove=${/* won't get executed */}
> >
>   <md-icon slot="trailing-icon">arrow_drop_down</md-icon>
> </md-filter-chip>
> ```

That is correct, the remove event will not fire with an entirely new custom trailing-icon slot. It would fire with a `remove-trailing-icon` slot.

vdegenne:
@asyncLiz Last attempt. I don't want to bother  I just studied the code thoroughly and tried different solutions, just making sure you fully understand this approach.
I renamed `removable` to `has-trailing`, `has-trailing` here basically  serve as a switch for `show the trailing icon` which default to the "remove". User can then use `trailing-icon` to change the appearance of the icon.
`get hasSlottedTrailingIcon()` is used to determine if a slot content was provided but only during click event which won't compromise SSR. But ideally why not just dispatching a global event instead so in the end we could end up with a more intuitive syntax, e.g.

```html
<md-filter-chip
  has-trailing
  @trailing-click=${this.#openChipMenu}
>
    <md-icon slot="trailing-icon">arrow_drop_down</md-icon>
    <!-- default slot -->
    <md-menu>...</md-menu>
</md-filter-chip>
```

Another advantage for this syntax is all the focus-ring navigation logic you wrote still act as intended and do not need additional code.

asyncLiz:
We use `has-*` attributes specifically for SSR workarounds, so it'd be nice to keep that consistent rather than adding functionality with one of them. In general, most users shouldn't need to add `has-*` attributes unless they start seeing a FOUC. One day we'll remove them entirely once CSS adds [`:has-slotted()`](https://github.com/WICG/webcomponents/issues/936).

 I'd prefer to keep `removable` so it's more explicit to users what feature they're adding. It's also a breaking change if we remove it, and we want to stay on and support 1.x for a while.

I think an event like `@trailing-click` and attribute like `trailing-icon-action` or a `slot="trailing-icon-button"` is likely the right approach, but I'd like some more planning before we add it. For now, let's scope this PR to just customizing the existing remove button's icon with a `slot="remove-trailing-icon"` so we can get it in faster.

vdegenne:
@asyncLiz Ok it's clear now, thanks for the heads-up. I'll complete this PR and probably leave the `trailing-icon` slot part to you because I'm not comfortable with the focus and aria logic yet.

chips/internal/_trailing-icon.scss
```diff
@@ -35,7 +35,7 @@
     width: var(--_icon-size);
   }
 
-  [name="trailing-icon"]::slotted(md-icon) {
+  [name="remove-trailing-icon"]::slotted(md-icon) {
      --md-icon-size: var(--_icon-size);
      color: inherit;
   }
```
asyncLiz, line 41: We should be able to remove all of this CSS by creating a wrapper `<span class="trailing icon">` around the slot. Then it will re-use shared `.icon` styles to size the icon:

```scss
.icon ::slotted(:first-child) {
  font-size: var(--_icon-size);
  height: var(--_icon-size);
  width: var(--_icon-size);
}
```

`color` inherits through slots, and `<md-icon>` explicitly has `color: inherit` so I don't think we need that declaration either.

asyncLiz (changes requested):
Looking great! Just a couple nits

chips/internal/_trailing-icon.scss
vdegenne: Nice, I just realize `font-size` on `md-icon` will size the icon to match this value, this is so perfect.

vdegenne:
Something worth bringing to your attention, when using a custom remove icon using `close` value to emulate the default appearance, results look different:

image
(top: default, bottom: using `close` icon)

asyncLiz:
> Something worth bringing to your attention, when using a custom remove icon using `close` value to emulate the default appearance, results look different:
> 
> image (top: default, bottom: using `close` icon)

Can you file a separate issue for that?

chips/internal/trailing-icons.ts
```diff
@@ -34,10 +34,14 @@ export function renderRemoveButton({
       @focus=${focusListener}>
       <md-focus-ring part="trailing-focus-ring"></md-focus-ring>
       <md-ripple ?disabled=${disabled}></md-ripple>
-      <svg class="trailing icon" viewBox="0 96 960 960" aria-hidden="true">
-        <path
-          d="m249 849-42-42 231-231-231-231 42-42 231 231 231-231 42 42-231 231 231 231-42 42-231-231-231 231Z" />
-      </svg>
+      <span class="trailing icon">
```
asyncLiz, line 37: this change is 🤌

asyncLiz (approved):

chips/internal/trailing-icons.ts
vdegenne: \ o / I like how it reverted to just that

PR #5352, opened 2023-12-27
fix(progress): progress buffer defaults to 0

### LINKED ISSUE #5251: Ignore Linear Progress' Buffer unless it is set?
### What is affected?

Component

### Description

Right now, `<md-linear-progress>` has its `buffer` value default to 1 and if we set `max` to any other value (usually 100 or the number of steps to process, number of files to copy for example), we have to set `buffer` to the same value as well. I think we should not have to worry about `buffer` value unless we actually want to and manually specify it.

While this would be a breaking change, I don't think anyone is having a special case where `buffer` should be at 1 without `max` is also 1.

### Reproduction

[Lit Playground](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy50cyIsImNvbnRlbnQiOiJpbXBvcnQgXCJAbWF0ZXJpYWwvd2ViL2FsbFwiOyJ9LHsibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPCFET0NUWVBFIGh0bWw-XG48aGVhZD5cbiAgPHNjcmlwdCB0eXBlPVwibW9kdWxlXCIgc3JjPVwiLi9zaW1wbGUtZ3JlZXRpbmcuanNcIj48L3NjcmlwdD5cbjwvaGVhZD5cbjxib2R5PlxuICA8bWQtbGluZWFyLXByb2dyZXNzIG1heD1cIjNcIiB2YWx1ZT1cIjBcIj48L21kLWxpbmVhci1wcm9ncmVzcz5cbjwvYm9keT5cbiJ9LHsibmFtZSI6InBhY2thZ2UuanNvbiIsImNvbnRlbnQiOiJ7XG4gIFwiZGVwZW5kZW5jaWVzXCI6IHtcbiAgICBcImxpdFwiOiBcIl4zLjAuMFwiLFxuICAgIFwiQGxpdC9yZWFjdGl2ZS1lbGVtZW50XCI6IFwiXjIuMC4wXCIsXG4gICAgXCJsaXQtZWxlbWVudFwiOiBcIl40LjAuMFwiLFxuICAgIFwibGl0LWh0bWxcIjogXCJeMy4wLjBcIlxuICB9XG59IiwiaGlkZGVuIjp0cnVlfV0) where a `max` value is set to 3 and now the buffer runs from 2 to 3:

image

### Workaround

If approved, the current source code can easily be changed to accomplish this:

- Change the `buffer` property to allow `undefined`:

```diff
export class LinearProgress extends Progress {
  /**
   * Buffer amount to display, a fraction between 0 and `max`.
   */
- @property({type: Number}) buffer = 1;
+ @property({type: Number}) buffer: number | undefined = undefined;
```

- Update `dotStyles`, the only place that uses `buffer` for now:

```diff
    const dotStyles = {
      transform: `scaleX(${
-       (this.indeterminate ? 1 : this.buffer / this.max) * 100
+       (this.indeterminate ? 1 : (this.buffer ?? this.max) / this.max) * 100
      }%)`,
    };
```

With the above code, adding and removing `buffer` attribute works as intended too: removing `buffer` HTML attribute sets `buffer` value to `undefined` and the buffer bar doesn't show up again.

### Is this a regression?

### Affected versions

1.0.1

N/A

datvm:
Fix #5251 .

```html
    <md-linear-progress value="10" max="100"></md-linear-progress>    
    <md-linear-progress value="10" buffer="50" max="100"></md-linear-progress>
```

image

### DIFF
--- a/progress/internal/linear-progress.ts
+++ b/progress/internal/linear-progress.ts
@@ -42,7 +42,7 @@ export class LinearProgress extends Progress {
     // Only display dots when visible - this prevents invisible infinite
     // animation.
     const hideDots =
-      this.indeterminate || !hasBuffer || this.value >= this.max;
+      this.indeterminate || !hasBuffer || bufferValue >= this.max || this.value >= this.max;
     return html`
       <div class="dots" ?hidden=${hideDots}></div>
       <div class="inactive-track" style=${styleMap(dotStyles)}></div>
--- a/progress/internal/linear-progress.ts
+++ b/progress/internal/linear-progress.ts
@@ -16,9 +16,9 @@ import {Progress} from './progress.js';
 export class LinearProgress extends Progress {
   /**
    * Buffer amount to display, a fraction between 0 and `max`.
-   * If negative, the buffer is not displayed.
+   * If the value is 0 or negative, the buffer is not displayed.
    */
-  @property({type: Number}) buffer = -1;
+  @property({type: Number}) buffer = 0;
 
   // Note, the indeterminate animation is rendered with transform %'s
   // Previously, this was optimized to use px calculated with the resizeObserver
@@ -29,17 +29,20 @@ export class LinearProgress extends Progress {
         (this.indeterminate ? 1 : this.value / this.max) * 100
       }%)`,
     };
+
+    const bufferValue = Math.min(this.buffer ?? 0, this.max);
+    const hasBuffer = bufferValue > 0;
+
+    const dotSize = this.indeterminate || !hasBuffer ? 1 : bufferValue / this.max;
+
     const dotStyles = {
-      transform: `scaleX(${
-        // == null is used to check for both null and undefined when buffer attribute is removed
-        (this.indeterminate || this.buffer == null || this.buffer < 0 ? 1 : this.buffer / this.max) * 100
-      }%)`,
+      transform: `scaleX(${dotSize * 100}%)`,
     };
 
     // Only display dots when visible - this prevents invisible infinite
     // animation.
     const hideDots =
-      this.indeterminate || this.buffer == null || this.buffer < 0 || this.buffer >= this.max || this.value >= this.max;
+      this.indeterminate || !hasBuffer || this.value >= this.max;
     return html`
       <div class="dots" ?hidden=${hideDots}></div>
       <div class="inactive-track" style=${styleMap(dotStyles)}></div>
--- a/progress/internal/linear-progress.ts
+++ b/progress/internal/linear-progress.ts
@@ -16,8 +16,9 @@ import {Progress} from './progress.js';
 export class LinearProgress extends Progress {
   /**
    * Buffer amount to display, a fraction between 0 and `max`.
+   * If negative, the buffer is not displayed.
    */
-  @property({type: Number}) buffer = 1;
+  @property({type: Number}) buffer = -1;
 
   // Note, the indeterminate animation is rendered with transform %'s
   // Previously, this was optimized to use px calculated with the resizeObserver
@@ -30,14 +31,14 @@ export class LinearProgress extends Progress {
     };
     const dotStyles = {
       transform: `scaleX(${
-        (this.indeterminate ? 1 : this.buffer / this.max) * 100
+        (this.indeterminate || this.buffer < 0 ? 1 : this.buffer / this.max) * 100
       }%)`,
     };
 
     // Only display dots when visible - this prevents invisible infinite
     // animation.
     const hideDots =
-      this.indeterminate || this.buffer >= this.max || this.value >= this.max;
+      this.indeterminate || this.buffer < 0 || this.buffer >= this.max || this.value >= this.max;
     return html`
       <div class="dots" ?hidden=${hideDots}></div>
       <div class="inactive-track" style=${styleMap(dotStyles)}></div>
--- a/progress/internal/linear-progress.ts
+++ b/progress/internal/linear-progress.ts
@@ -31,14 +31,15 @@ export class LinearProgress extends Progress {
     };
     const dotStyles = {
       transform: `scaleX(${
-        (this.indeterminate || this.buffer < 0 ? 1 : this.buffer / this.max) * 100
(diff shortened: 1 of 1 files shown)


progress/internal/linear-progress.ts
```diff
@@ -30,14 +31,14 @@ export class LinearProgress extends Progress {
     };
     const dotStyles = {
       transform: `scaleX(${
-        (this.indeterminate ? 1 : this.buffer / this.max) * 100
+        (this.indeterminate || this.buffer < 0 ? 1 : this.buffer / this.max) * 100
```
asyncLiz, line 34: Does `this.buffer <= 0` handle the null case correctly?

When `null`, it'll coerce to `0`. If a buffer is `0`, it won't display anyway.

datvm: @asyncLiz unfortunately not. `dotStyles` is evaluated to `scaleX(0%)` but `hideDots` is still evaluated to `false` so the whole buffer shows up after removing it:

image

image

I am making another commit to fix it soon.

datvm: New commit should now correctly handle removing `buffer` attribute. I am not sure if it's better to set its type to `number | null` because it reflects reality better and possibly saves us from some unintended bug in the future. Nevertheless you are more aware of the whole build process than me.

datvm:
@asyncLiz hi, anything else for this PR?

progress/internal/linear-progress.ts
```diff
@@ -30,14 +31,15 @@ export class LinearProgress extends Progress {
     };
     const dotStyles = {
       transform: `scaleX(${
-        (this.indeterminate ? 1 : this.buffer / this.max) * 100
+        // == null is used to check for both null and undefined when buffer attribute is removed
+        (this.indeterminate || this.buffer == null || this.buffer < 0 ? 1 : this.buffer / this.max) * 100
```
asyncLiz, line 35: This is starting to get hard to grok, let's split it out and simplify it.

```ts
// use ?? for null and undefined when the buffer attribute is removed.
const bufferValue = this.buffer ?? 0;
const hasBuffer buffer > 0;
const dotSize = this.indeterminate || !hasBuffer ? 1 : bufferValue / this.max;
const dotStyles = {
  transform: `scaleX(${dotSize * 100}%)`
};
```

progress/internal/linear-progress.ts
```diff
@@ -30,14 +31,15 @@ export class LinearProgress extends Progress {
...
       }%)`,
     };
 
     // Only display dots when visible - this prevents invisible infinite
     // animation.
     const hideDots =
-      this.indeterminate || this.buffer >= this.max || this.value >= this.max;
+      this.indeterminate || this.buffer == null || this.buffer < 0 || this.buffer >= this.max || this.value >= this.max;
```
asyncLiz, line 42: And use the new variables here

```ts
const hideDots =
  this.indeterminate || !hasBuffer || bufferValue >= this.max || this.value >= this.max;
```

progress/internal/linear-progress.ts
```diff
@@ -16,8 +16,9 @@ import {Progress} from './progress.js';
 export class LinearProgress extends Progress {
   /**
    * Buffer amount to display, a fraction between 0 and `max`.
+   * If negative, the buffer is not displayed.
```
asyncLiz, line 19: Hm, actually what about using `0` as the default value to indicate that there is no buffer? I think that makes more better sense than using negative to indicate it.

If there's 0 buffer after all, there is no buffer.

asyncLiz (changes requested):
Thank you for the ping! I lost track of this.

progress/internal/linear-progress.ts
datvm: @asyncLiz thanks, the latest commit should address the above suggestions

progress/internal/linear-progress.ts
```diff
@@ -28,16 +29,20 @@ export class LinearProgress extends Progress {
         (this.indeterminate ? 1 : this.value / this.max) * 100
       }%)`,
     };
+
+    const bufferValue = Math.min(this.buffer ?? 0, this.max);
```
asyncLiz, line 33: I'm not sure I entirely follow why we need `Math.min` here to change bufferValue to max. Is it to avoid `bufferValue >= this.max` for `const hideDots`?

datvm: Yes, so we do not need to worry about that logic with `dotStyles` anymore though I suppose the `hideDots` would take care of it anyway right?

asyncLiz: Yea I believe so. Do you mind not clamping it and keeping the `bufferValue >= this.max` check?

Semantically "bufferValue" as changed doesn't reflect `this.buffer`, it's really `bufferValueOrMax`. It's a bit confusing to read, and may inadvertently cause problems if future changes don't realize that nuance.

datvm: @asyncLiz sure I did in the last commit.

progress/internal/linear-progress.ts
```diff
@@ -28,16 +29,20 @@ export class LinearProgress extends Progress {
...
-        (this.indeterminate ? 1 : this.buffer / this.max) * 100
-      }%)`,
+      transform: `scaleX(${dotSize * 100}%)`,
     };
 
     // Only display dots when visible - this prevents invisible infinite
     // animation.
     const hideDots =
-      this.indeterminate || this.buffer >= this.max || this.value >= this.max;
+      this.indeterminate || !hasBuffer || this.value >= this.max;
```
asyncLiz, line 45: Last change I think! We still need the max check, `!hasBuffer || bufferValue >= this.max`

asyncLiz (changes requested):

progress/internal/linear-progress.ts
datvm: Oops 😬 sorry I forgot about that one after removing the previous one. It should be updated now.

asyncLiz (approved):

PR #5363, opened 2024-01-03
fix(button): add missing sass imports

jason2506:


asyncLiz:
I don't believe these files use Sass's list module, so the imports are not needed

asyncLiz (approved):
Ohhh gotcha, thank you for catching this!

asyncLiz:
Just need you to take a look at the [CLA](https://github.com/material-components/material-web/pull/5363/checks?check_run_id=20123813706) and then we can import :)

PR #5399, opened 2024-01-19
fix(testing): add testing folder to npm package

### LINKED ISSUE #5397: Add harness to npm package
### Description

Hi, 
I am currently working on a Lit integration for Tanstack Form. I am currently in the process of writing tests and sadly there is no good framework for writing web component tests. 

I have opened a couple testing library PRs but they haven't been merged in ages: https://github.com/testing-library/user-event/pull/1033

Hence, I am looking for a different way to simulate input in my tests. Since I am using Material Web for components I thought maybe I could use the `Harness`: https://github.com/material-components/material-web/blob/main/testing/harness.ts Sadly, it seems harness is not part of the npm package however. 

Is there a reason for this? Angular Material does provide component harnesses and I would welcome if Material web did too.

Christian24:
This should fix #5397.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,11 +43,13 @@
     "!web-test-runner.config.js",
     "!commitlint.config.js",
     "!**/test/**",
-    "!**/testing/**",
     "!**/*_test.*",
     "!.wireit/**",
     "!catalog",
-    "!scripts/"
+    "!scripts/",
+    "!**/testing/**",
+    "testing/harness.{js,js.map,d.ts}",
+    "testing/transform-pseudo-classes.{js,js.map,d.ts}"
   ],
   "workspaces": [
     "catalog"


asyncLiz (approved):
LGTM, just need to double check the package contents before importing

package.json
```diff
@@ -43,7 +43,6 @@
     "!web-test-runner.config.js",
     "!commitlint.config.js",
     "!**/test/**",
-    "!**/testing/**",
```
asyncLiz, line 46: Let's keep excluding all the other testing files and explicitly add the files that we consider to be part of the public API for harnesses that clients can use in their tests:

```
    "!**/test/**",
    "!**/testing/**",
    "testing/harness.{js,js.map,d.ts}",
    "testing/transform-pseudo-classes.{js,js.map,d.ts}",
```

asyncLiz (changes requested):
one small change first!

package.json
Christian24: I added that :)

asyncLiz (approved):

package.json
```diff
@@ -87,7 +86,10 @@
         "!**/*.css.ts",
         "!catalog",
         "!scripts/",
-        "!node_modules"
+        "!node_modules",
+        "!**/testing/**",
```
asyncLiz, line 90: I think this was added to the wrong section. It should be in the top level "files" on line 46, not the build wireit command

asyncLiz (changes requested):

package.json
Christian24: Oops. Fix incoming

asyncLiz (approved):

Christian24:
I reworded the commit message so hopefully it passes listing.

asyncLiz:
> I reworded the commit message so hopefully it passes listing.

_aggressively stamps each import email change_ 😂

No worries, it'll use the pr title when merging. Sometimes the check gets stale :)

Christian24:
It seems it is still not happy with the PR title...

asyncLiz:
> It seems it is still not happy with the PR title...

Nw, it'll get all fixed later! Just waiting on Elliott to stamp it internally and the bot will do the rest :P

asyncLiz:
> > It seems it is still not happy with the PR title...
> 
> Nw, it'll get all fixed later! Just waiting on Elliott to stamp it internally and the bot will do the rest :P

Ohhhh wait it's a type that it doesn't know about. Lemme fix that.

asyncLiz:
It should be good to go after https://github.com/material-components/material-web/pull/5410

PR #5444, opened 2024-02-02
feat(catalog): Added fixed colors to the palette

vdegenne:
If you ever need to use fixed colors in some parts of the catalog, this PR adds them to the palette.
It also sorts and prettify the keys of the `materialColors` object.


vdegenne:
I will never get the titles right..

asyncLiz:
> I will never get the titles right..

haha nw, sorry 😅

The catalog isn't really a "feature" of `@material/web`, but it'd be nice to call it out. I wish there was a "chore" or "note"/"misc" section for the changelog.

PR #5448, opened 2024-02-07
fix(buttons): Disable pointer-events on background & touch

VandeurenGlenn:


asyncLiz:
The purpose of the touch target is to receive pointer events to extend the interactive surface of an element for accessibility, so we can't turn them off.

VandeurenGlenn:
Closing this since the actual issue was a wrong listener target

Issue #5398, opened 2024-01-19
Radio throws error and hydrates tabindex in some hydration scenarios


einarstubhaug:
### What is affected?

Component

### Description

When I try to use radio buttons it fails with error messages.
In Chrome:  _Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes_
Firefox: Uncaught DOMException: Operation is not supported

### Reproduction

https://stackblitz.com/edit/webpack-webpack-js-org-slhiao?file=dist%2Findex.html

### Workaround

I have not found a workaround

### Is this a regression?

Yes. This used to work, but now it doesn't.

### Affected versions

Failing in 1.1.1

### Browser/OS/Node environment

Latest Chrome, Mac Sonoma, Node 20.10.0

asyncLiz:
Hmm interesting! Ultimately this is failing because of a `tabindex` attribute. We set the tabindex after [`queueMicrotask()`](https://github.com/material-components/material-web/blob/4b4c373a6b9ff28f4acf604335701228525c3b84/labs/behaviors/focusable.ts#L80-L83) to try and set it as soon as possible after construction.

This may not be long enough in some situations. We could use a `setTimeout()` to wait a full task. When I update your compiled `main.js` and replace `queueMicrotask` with `setTimeout`, the radios appear.

Alternatively, we could refactor a bit so that we don't update the `tabindex` attribute until the element is first connected. That's probably the better approach, but a bit more effort.

Would welcome a PR to swap it to setTimeout in the meantime!

asyncLiz:
Fixed in https://github.com/material-components/material-web/pull/5463

PR #5463, opened 2024-02-14
fix: tabindex causing error that elements like radio must not have attributes

CarlosEGuerraSilva:
Fix for #5398 , when hydrating tabIndex on radios and tabs, it causes a **DOMException**. The proposed solution suggests moving the `hasConstructed` property and calling the `updateTabIndex` method in the `connectedCallback` to be able to modify the _tabIndex_ as soon as the element is connected to the **DOM** instead using `setTimeout`


labs/behaviors/focusable.ts
```diff
@@ -77,10 +77,12 @@ export function mixinFocusable<T extends MixinBase<LitElement>>(
     constructor(...args: any[]) {
       super(...args);
       this[isFocusable] = true;
-      queueMicrotask(() => {
-        this[hasConstructed] = true;
-        this[updateTabIndex]();
-      });
+    }
+
+    connectedCallback() {
```
asyncLiz, line 82: Full disclosure: this may be a harder change to land. Changing timings from a microtask to a task could be hard enough with test timings to consider. Moving to another lifecycle entirely might result in more breakages.

Won't know until we do some more internal testing though

labs/behaviors/focusable.ts
```diff
@@ -77,10 +77,12 @@ export function mixinFocusable<T extends MixinBase<LitElement>>(
...
+      super.connectedCallback();
+      this[hasConstructed] = true;
```
asyncLiz, line 84: I think we can completely remove the `hasConstructed` property, symbol, and the constructor by changing `privateIsFocusable` to default to false.

```ts
    [privateIsFocusable] = true;
    [externalTabIndex]: number | null = null;
    [isUpdatingTabIndex] = false;

    // * also don't forget `override`!
    override connectedCallback() {
      super.connectedCallback();
      this[updateTabIndex]();
    }
```

We needed `hasConstructed` because we set `isFocusable` in the constructor, which called `updateTabIndex`.

If we bypass `isFocusable = true` and directly set `privateIsFocusable = true`, then we don't need to do that dance.

asyncLiz (changes requested):

labs/behaviors/focusable.ts
CarlosEGuerraSilva: Seems fine to me, as such having Microtask or setTimeout can end up too early or too short giving timing problems anyway in all sorts of scenarios.

labs/behaviors/focusable.ts
```diff
@@ -68,19 +67,19 @@ export function mixinFocusable<T extends MixinBase<LitElement>>(
       this[updateTabIndex]();
     }
 
-    [privateIsFocusable] = false;
+    [privateIsFocusable] = true;
     [externalTabIndex]: number | null = null;
     [isUpdatingTabIndex] = false;
-    [hasConstructed] = false;
 
     // tslint:disable-next-line:no-any
```
asyncLiz, line 74: We can remove the constructor entirely. `this[isFocusable] = true` won't do anything the `[privateIsFocusable] = true` as the default.

asyncLiz (changes requested):

labs/behaviors/focusable.ts
CarlosEGuerraSilva: Yup, after removing keeps working

asyncLiz (approved):

asyncLiz:
Gonna need to do some internal testing on this, thanks for your patience!

PR #5492, opened 2024-02-26
fix: `required` attribute is not passed to underlying input element

CornerSyrup:


asyncLiz:
What problem is this PR trying to solve? The underlying input is not used for form association, so adding required to it shouldn't change anything.

CornerSyrup:
I apologise for wasting your time... I was using older version of MWC and the `required` problem probably solved before...

PR #5501, opened 2024-02-28
fix(switch): `selected` state is updated during input event

vdegenne:
Fixes https://github.com/material-components/material-web/issues/5486

### DIFF
--- a/switch/internal/switch.ts
+++ b/switch/internal/switch.ts
@@ -114,6 +114,7 @@ export class Switch extends switchBaseClass {
           ?checked=${this.selected}
           ?disabled=${this.disabled}
           ?required=${this.required}
+          @input=${this.handleInput}
           @change=${this.handleChange} />
 
         <md-focus-ring part="focus-ring" for="switch"></md-focus-ring>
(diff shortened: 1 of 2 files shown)


asyncLiz:
This is intentional, the `selected` attribute represents the [default state of the form control](https://github.com/material-components/material-web/blob/274ce3e4e06867acfab642123192e0a7aa9b45d4/switch/internal/switch.ts#L211-L215), not its current active state. This matches `<input type="checkbox" checked>` behavior, whether the `checked` attribute does not change when the value changes.

https://github.com/material-components/material-web/issues/5486 should be fixed by changing the switch to update its state on `'input'` [instead of `'change'`](https://github.com/material-components/material-web/blob/274ce3e4e06867acfab642123192e0a7aa9b45d4/switch/internal/switch.ts#L195), [similar to checkbox](https://github.com/material-components/material-web/blob/274ce3e4e06867acfab642123192e0a7aa9b45d4/checkbox/internal/checkbox.ts#L171-L181).

vdegenne:
@asyncLiz Sorry to tag you but I couldn't request a review you were not automatically placed in the reviewers list.

That should do it

asyncLiz (approved):
Looks great!

PR #5521, opened 2024-03-09
fix(slider): support linear-gradient track background colors

### LINKED ISSUE #5518: md-slider track background
### What is affected?

Theming

### Description

Hi, I want to set slider tick's color tolinear-gradient with setting --_active-track-color variable to gradient. But it doesn't work because the variable sets background-color of tick instead of background. background-color doesnt support linear gradient.
Is there a way to bypass or fix it?

### Reproduction

 It doesn't work.

### Workaround

I have not found a workaround.

### Is this a regression?

### Affected versions

1.0.3

All platforms

TeamBusylj:
Fixes #5518

I have changed md-slider track's css property background-color to background, so now it is possible to set a gradient as bg of slider's track. Example: 

image

### DIFF
--- a/slider/internal/_slider.scss
+++ b/slider/internal/_slider.scss
@@ -163,7 +163,7 @@ $_md-sys-shape: tokens.md-sys-shape-values();
   }
 
   .track::before {
-    background-color: var(--_inactive-track-color);
+    background: var(--_inactive-track-color);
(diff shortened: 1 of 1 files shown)


asyncLiz (approved):
Thanks for this fix!

PR #5533, opened 2024-03-18
fix(iconbutton): handle activation click

vdegenne:
fixes https://github.com/material-components/material-web/issues/5532

### DIFF
--- a/iconbutton/internal/icon-button.ts
+++ b/iconbutton/internal/icon-button.ts
@@ -7,8 +7,8 @@
 import '../../focus/md-focus-ring.js';
 import '../../ripple/ripple.js';
 
-import {html, LitElement, nothing} from 'lit';
-import {property, state} from 'lit/decorators.js';
+import {html, isServer, LitElement, nothing} from 'lit';
+import {property, query, state} from 'lit/decorators.js';
 import {classMap} from 'lit/directives/class-map.js';
 import {literal, html as staticHtml} from 'lit/static-html.js';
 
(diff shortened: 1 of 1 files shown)


vdegenne:
@asyncLiz I followed the same idea as `<md-text-button>`, however to be totally honest I don't know if that's good or not. Requesting your review.

asyncLiz:
The form label activation is used to make `<label>` work with some wrapped elements like `<input>`. I think it's fixing the issue as a side effect by forwarding clicks.

Can we try overriding the click method instead? I think it makes more sense to forward the initial click to the delegated element, rather than trigger a host click which re-triggers a button click.

```ts
override click() {
  this.buttonElement?.click();
}
```

If it's more complicated then I'm open to the current approach, but I want to add some documentation to explain what's going on.

iconbutton/internal/icon-button.ts
```diff
@@ -126,6 +130,15 @@ export class IconButton extends iconButtonBaseClass implements FormSubmitter {
 
   @state() private flipIcon = isRtl(this, this.flipIconInRtl);
 
+  @query('.link') private readonly linkElement!: HTMLElement | null;
```
asyncLiz, line 133: Update this to capture both the button and anchor element that could be rendered. Both can have clicks forwarded to them.

```ts
@query('.icon-button') private readonly buttonElement!: HTMLElement | null;
```

asyncLiz (changes requested):

PR #5563, opened 2024-04-03
fix: buttons' appearance in disabled fieldset

datvm:
Fix #5562 

image

PR #5565, opened 2024-04-03
fix: formDisabledCallback and <fieldset> disabled attribute

### LINKED ISSUE #5049: use formDisabledCallback to support <fieldset> disabled attribute
### What is affected?

Component

### Description

See: https://jsfiddle.net/datvm/zdpeqc5j/1/

image

For standard components, when an acestor `<fieldset>` has `[disabled=true]`, they are disabled as well. I think this behavior is not discussed yet so it's probably not a bug? Would you consider adding this feature?

Note: the screenshot above misses `<md-select>`.

### Reproduction

https://jsfiddle.net/datvm/zdpeqc5j/1/

```html
<fieldset disabled>
    <p>
        <md-filled-button>A Material 3 Button</md-filled-button>
        <button>Standard button</button>
    </p>
    
    <p>
        <md-outlined-text-field></md-outlined-text-field>
        <input value="Standard input" />
    </p>
    
    <p>
        <md-checkbox></md-checkbox>
        <md-switch></md-switch>
        <input type="checkbox" value="Standard input" />
    </p>
    
    <p>
        <md-radio></md-radio>
        <input type="radio" value="Standard input" />
    </p>
    
    <p>
        <md-slider></md-slider>
        <input type="range" />
    </p>
    
</fieldset>
```

### Workaround

You have to manually set `disabled` to each component in the `fieldset`.

### Is this a regression?

### Affected versions

1.0.0

-

datvm:
Fixes #5049 where the control `disabled` property itself is changed which would cause issues for `fieldset`.

This commit need to be merged along with #5563. I am sorry I am not familiar with the process if one PR is depend on the other. In that commit, the render/UI/behavior make use of `:disabled` instead of the `disabled` property only, therefore the behavior should matches with native elements:

https://github.com/material-components/material-web/assets/6388546/7ca89b96-1d7e-4bd8-8dfc-1d901904fadc

### DIFF
--- a/button/internal/_elevation.scss
+++ b/button/internal/_elevation.scss
@@ -20,7 +20,7 @@ $_md-sys-motion: tokens.md-sys-motion-values();
     transition-timing-function: map.get($_md-sys-motion, 'emphasized-easing');
   }
 
-  :host([disabled]) md-elevation {
+  :host(:disabled) md-elevation {
     transition: none;
   }
 
@@ -59,7 +59,7 @@ $_md-sys-motion: tokens.md-sys-motion-values();
     );
   }
 
-  :host([disabled]) md-elevation {
+  :host(:disabled) md-elevation {
     @include elevation.theme(
       (
         'level': var(--_disabled-container-elevation),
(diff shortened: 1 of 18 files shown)


asyncLiz:
Let's wait for https://github.com/material-components/material-web/pull/5563 to land, then you can sync to head for it.

PR #5588, opened 2024-04-17
docs: fix typo on theming docs

aarongarciah:
There aren't CSS vars prefixed with `--md-sys--typograph`. The correct prefix is `--md-sys--typescale`. See https://github.com/material-components/material-web/blob/main/tokens/_md-sys-typescale.scss#L138.

The "Typography" heading could also be renamed to "Typescale" for consistency with how the section is named in the [Typography page](https://material-web.dev/theming/typography/#typescale).

### DIFF
--- a/docs/theming/README.md
+++ b/docs/theming/README.md
@@ -88,14 +88,14 @@ to components. See the [color guide](color.md) for more details.
 
 #### Typography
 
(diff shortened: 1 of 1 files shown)


asyncLiz (approved):
Thanks for the fix!

Issue #5608, opened 2024-05-03
Limitations of <md-tabs>, missing vertical option.


Dakotys:
### Description

I feel like <md-tabs> is missing an option to display the tabs vertically to display them on side of the page.

asyncLiz:
Thanks for the issue! I mentioned in the PR that this is the [Material navigation rail component](https://m3.material.io/components/navigation-rail/overview), which we haven't built yet. I'll update the title to track the implementation.

coteyka1209:
+1, i need navigation rail so much

PR #5609, opened 2024-05-03
feat: vertical support for tabs component

Dakotys:
I felt like this was missing from tabs component and decided to add possible implementation of mentioned functionality myself. 
To use it just add vertical attribute to `<md-tabs>` and they are displayed vertically with indicator staying on bottom of the individual tabs.
I also added vertical attribute to `<md-primary-tab>` and `<md-secondary-tab>` so the indicator can be put sideways.
The only missing thing is maybe updated documentation.
#5608

### DIFF
--- a/tabs/internal/_tab.scss
+++ b/tabs/internal/_tab.scss
@@ -202,7 +202,10 @@
     .content{
       width: 100%;
       padding: 0 16px;
-    }
+
+      &.has-icon{
+        justify-content: start;
+      }
 
     .indicator {
       inset: 0 0 0 auto;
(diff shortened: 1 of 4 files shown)


asyncLiz:
Thank you for your contribution! Unfortunately though, this doesn't align with Material's [guidelines for tab](https://m3.material.io/components/tabs/overview).

The component pattern I think you're looking for is a [navigation rail](https://m3.material.io/components/navigation-rail/overview). We haven't built that one yet, but it'd be much more involved than making the current tabs vertical.

While we can't accept this, you can always `class MyVerticalTab extends MdTab` to add this functionality for your app!

PR #5616, opened 2024-05-14
feat(field): make required asterisk optional

### LINKED ISSUE #5256: Text field required asterisk should be optional
### What is affected?

Component

### Description

According to the md3 docs the required asterisk on the label text should not be present once the field has a value:
https://m3.material.io/components/text-fields/guidelines#9ad90554-a793-4506-9075-6812fd7b381a

There should also be a way to choose if you want the asterisk to show at all, as per https://m3.material.io/components/text-fields/guidelines#6c36a812-273d-4f5d-9102-2f00386c9145 for when you want to highlight the optional fields and not required ones.- I couldn't see a way to do this in docs.

### Reproduction

N/A

### Workaround

N/A

### Is this a regression?

### Affected versions

Nightly

Edge Version 119.0.2151.97 (Official build) (64-bit)

npeters-dev:
Fixes #5256

Makes the required asterisks optional on text fields by adding an `no-asterisk` prop.


asyncLiz (changes requested):
Looks great! Do you mind adding the same property to the select component?

npeters-dev:
> Looks great! Do you mind adding the same property to the select component?

Not at all. I added the property to the select component, too.

asyncLiz (approved):
Awesome work!

asyncLiz:
I think more commits got pushed to the PR before copybara finished the import, so it didn't merge propertly. Sorry about that! This got added in https://github.com/material-components/material-web/commit/c5e87b23a100800b9b89c733a35f5edbc6ea38a2 though, so we can close the PR now :)

Issue #4589, opened 2023-07-21
Text Field components do not support name attribute


caleb-sideras:
The md-[filled/outlined]-text-field components do not give you access to the underlying "name" input field attribute; so cannot use this component in a form without using JS. It's a bit unusual that these components do not support the name attribute, as it is pretty standard and would make these components more versatile and compatible with traditional form handling. 

The example below does not post the name attribute:
```html
<md-filled-text-field label="Add a new task" type="text" name="task"></md-filled-text-field>
```

asyncLiz:
I believe [text field has a `name` property](https://github.com/material-components/material-web/blob/b5712f3d4b4845e98cfb74f145e400647b492823/textfield/lib/text-field.ts#L124-L132) and should work in form submission.

Example: https://lit.dev/playground/#gist=4bacf0a4b874c4a825f943bf8d0d946a

caleb-sideras:
You're right thanks, rookie mistake. Couldn't see the name attribute in docs/source nor on the input element in DOM.

engrbm87:
I noticed that the name attribute is not added to the input element. This disables autocomplete in the browser. The above example uses `form` but in case a form is not used the autocomplete doesn't work if the name attribute is not added to the input element.

asyncLiz:
> I noticed that the name attribute is not added to the input element. This disables autocomplete in the browser. The above example uses `form` but in case a form is not used the autocomplete doesn't work if the name attribute is not added to the input element.

Do you mind opening a PR to forward the name attribute onto the inner `<input>` and `<textarea>` elements to support browser inferring autocomplete outside of a form?

PR #5617, opened 2024-05-15
fix(textfield): browser autocomplete works outside of a form

engrbm87:
Add name attribute to input and textarea elements created by the text-filed element. 
This enables autocomplete to work int he browser even if a form is not used.
Discussed in issue #4589

### DIFF
--- a/textfield/internal/text-field.ts
+++ b/textfield/internal/text-field.ts
@@ -604,6 +604,7 @@ export abstract class TextField extends textFieldBaseClass {
           aria-invalid=${this.hasError}
           aria-label=${ariaLabel}
           autocomplete=${autocomplete || nothing}
+          name=${this.name || nothing}
           ?disabled=${this.disabled}
           maxlength=${hasMaxLength ? this.maxLength : nothing}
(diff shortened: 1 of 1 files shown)


textfield/internal/text-field.ts
```diff
@@ -604,6 +609,7 @@ export abstract class TextField extends textFieldBaseClass {
           aria-invalid=${this.hasError}
           aria-label=${ariaLabel}
           autocomplete=${autocomplete || nothing}
+          name=${this.name}
```
asyncLiz, line 612: Can you add `|| nothing` to not render the name attribute when none is provided?

```
name=${this.name || nothing}
```

asyncLiz (changes requested):
Can you revert all formatting changes? Try not to run a formatter like prettier on our files, since it may not match our internal formatting requirements.

Thanks!

textfield/internal/text-field.ts
engrbm87: Thanks for the comments. PR updated as advised.

asyncLiz (approved):
Awesome!

PR #5635, opened 2024-06-05
docs: update analyzer and update docs

e111077:
The analyzer now works. This means our API docs will now auto-sync with updates to jsdoc


asyncLiz (approved):
yesssssss <3

PR #5663, opened 2024-06-26
refactor(menu): focus before animation

### LINKED ISSUE #5662: Menu should focus during opening animation
### What is affected?

Accessibility, Component

### Description

Currently when opening a menu, the entire opening animation has to finish before the default focus is applied, which is rather annoying as a keyboard user.

### Reproduction

https://material-web.dev/components/menu/

### Workaround

The only workaround is setting quick open on all your menus to skip the animation.

### Is this a regression?

### Affected versions

1.5.1

Browser: Edge 126.0.2592.68
OS: Windows 11
Node version: 22.3.0
npm version: 10.8.1

austinw-fineart:
fixes #5662

### DIFF
--- a/menu/internal/menu.ts
+++ b/menu/internal/menu.ts
@@ -596,21 +596,15 @@ export abstract class Menu extends LitElement {
 
     let animationAborted = !this.quick;
 
-    if (this.quick) {
-      this.dispatchEvent(new Event('opening'));
-    } else {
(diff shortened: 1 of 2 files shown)


asyncLiz:
Thanks for the PR! At a glance I don't see anything problematic with this approach, but it may take me a little bit to import and test

PR #5681, opened 2024-07-15
feat: check for server before creating treewalker from document

SaadBazaz:
Hey, 

I've been trying to get Material Components to work in NextJS SSG. All other components work fine [except for Dialog](https://github.com/grayhatdevelopers/material-web-react/blob/main/apps/demo/src/app/page.tsx#L14), which throws the error `document is not defined`.

Upon inspection, the reason is because Dialog uses document.treeWalker upon initialization. I've added a check to initialize it as an empty object typecasted to TreeWalker, so that it works fine in SSG.

Tested, works ok. However, I'd love to know if there's a better way, instead of typecasting.

### DIFF
--- a/dialog/internal/dialog.ts
+++ b/dialog/internal/dialog.ts
@@ -152,7 +152,7 @@ export class Dialog extends dialogBaseClass {
   // focusable elements. TreeWalker is faster than `querySelectorAll('*')`.
   // We check for isServer because there isn't a "document" during an SSR
   // run.
-  private readonly treewalker = isServer ? {} : document.createTreeWalker(
+  private readonly treewalker = isServer ? {} as TreeWalker : document.createTreeWalker(
     this,
     NodeFilter.SHOW_ELEMENT,
   );
--- a/dialog/internal/dialog.ts
+++ b/dialog/internal/dialog.ts
@@ -150,7 +150,9 @@ export class Dialog extends dialogBaseClass {
   private escapePressedWithoutCancel = false;
   // This TreeWalker is used to walk through a dialog's children to find
   // focusable elements. TreeWalker is faster than `querySelectorAll('*')`.
-  private readonly treewalker = document.createTreeWalker(
+  // We check for isServer because there isn't a "document" during an SSR
+  // run.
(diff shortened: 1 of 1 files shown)


dialog/internal/dialog.ts
```diff
@@ -150,7 +150,9 @@ export class Dialog extends dialogBaseClass {
   private escapePressedWithoutCancel = false;
   // This TreeWalker is used to walk through a dialog's children to find
   // focusable elements. TreeWalker is faster than `querySelectorAll('*')`.
-  private readonly treewalker = document.createTreeWalker(
+  // We check for isServer because there isn't a "document" during an SSR
+  // run.
+  private readonly treewalker = isServer ? {} as TreeWalker : document.createTreeWalker(
```
asyncLiz, line 155: Rather than setting it to an empty object, which could cause unexpected type failures, change the type to `TreeWalker|null`

```ts
private readonly treewalker = isServer ? null : document.createTreeWalker(...)
```

Then, check for null in `getFirstAndLastFocusableChildren()` and return early. This is more type-safe than `{} as TreeWalker`.

```ts
  private getFirstAndLastFocusableChildren() {
    if (!this.treewalker) {
      return [null, null] as const;
    }

    let firstFocusableChild: HTMLElement | null = null;
    let lastFocusableChild: HTMLElement | null = null;
```

The `as const` tells the return type to be a strict `[null, null]` tuple rather than an array of unknown length.

asyncLiz (changes requested):
Good catch! One small change request to have stricter types

dialog/internal/dialog.ts
SaadBazaz: Thanks for the detailed feedback, @asyncLiz.
I've added the null ternary, added the nullcheck, and removed the typecasts.
Instead of typecasting as const, I've added an explicit return type to the function, so that it can help wherever it's used.

asyncLiz: That sounds fine to me!

asyncLiz (approved):
Thank you!

PR #5700, opened 2024-08-23
fix(radio): change touch target to fix blocked aria-hidden on an <input> element error

### LINKED ISSUE #5699: Accessibility issue with aria-hidden="true" attribute in <md-radio> component
### What is affected?

Accessibility

### Description

When clicking the button part of the radio button in the Shadow DOM of the <md-radio> component, the following error occurs:

```
Blocked aria-hidden on a <input> element because the element that just received focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden. 
```

This is caused by the aria-hidden="true" attribute assigned to the button part of the radio button, which hides the focused element from accessibility tools.

To solve this problem, we propose replacing the aria-hidden attribute with the inert attribute. The inert attribute prevents focus as well as hides the element from accessibility tools.

### Reproduction

https://stackblitz.com/edit/nuxt-starter-wdxt3x?file=app.vue

### Workaround

Instead of aria-hidden=“true”, use the inert attribute.

### Is this a regression?

### Affected versions

2.0.0

Browser: Chrome 127.0.6533.120
OS: macOS 14.6.1
Node version: 18.20.3
npm version: 10.2.3

pennin-style:
Fixes #5699

### DIFF
--- a/radio/internal/_radio.scss
+++ b/radio/internal/_radio.scss
@@ -89,17 +89,13 @@ $_md-sys-motion: tokens.md-sys-motion-values();
       );
     }
 
-    // <input> is also the touch target
-    input {
-      appearance: none;
+    .touch-target {
       height: 48px;
-      margin: 0;
       position: absolute;
       width: 48px;
-      cursor: inherit;
     }
 
-    :host([touch-target='none']) input {
+    :host([touch-target='none']) .touch-target {
       width: 100%;
       height: 100%;
     }
--- a/radio/internal/_radio.scss
+++ b/radio/internal/_radio.scss
@@ -96,8 +96,7 @@ $_md-sys-motion: tokens.md-sys-motion-values();
     }
 
     :host([touch-target='none']) .touch-target {
-      width: 100%;
-      height: 100%;
+      display: none;
     }
 
     md-ripple {
--- a/radio/internal/radio.ts
+++ b/radio/internal/radio.ts
@@ -99,7 +99,7 @@ export class Radio extends radioBaseClass {
   protected override render() {
     const classes = {'checked': this.checked};
     return html`
-      <div class="container ${classMap(classes)}" aria-hidden="true">
+      <div class="container ${classMap(classes)}" inert">
         <md-ripple
           part="ripple"
           .control=${this}
--- a/radio/internal/radio.ts
+++ b/radio/internal/radio.ts
@@ -99,7 +99,7 @@ export class Radio extends radioBaseClass {
   protected override render() {
     const classes = {'checked': this.checked};
     return html`
-      <div class="container ${classMap(classes)}" inert">
+      <div class="container ${classMap(classes)}" aria-hidden="true">
         <md-ripple
           part="ripple"
           .control=${this}
@@ -119,13 +119,7 @@ export class Radio extends radioBaseClass {
           <circle class="inner circle" cx="10" cy="10" r="5" />
         </svg>
 
-        <input
-          id="input"
-          type="radio"
-          tabindex="-1"
-          .checked=${this.checked}
-          .value=${this.value}
-          ?disabled=${this.disabled} />
+        <div class="touch-target"></div>
       </div>
     `;
   }


asyncLiz:
Mentioned in the PR, but we still need a touch target that isn't `inert`. I think we can remove the input entirely and replace it with a touch target. Can you update this PR for that?

Replace [the `<input>` here](
https://github.com/material-components/material-web/blob/024d6d93cc0f83fb5540ea2bba21ecda2ba4d3f0/radio/internal/radio.ts#L122) with:

```html
<div class="touch-target"></div>
```

Replace the [`<input>` touch target styles here](https://github.com/material-components/material-web/blob/024d6d93cc0f83fb5540ea2bba21ecda2ba4d3f0/radio/internal/_radio.scss#L92-L105) with:

```scss
    .touch-target {
      height: 48px;
      position: absolute;
      width: 48px;
    }

    :host([touch-target='none']) .touch-target {
      width: 100%;
      height: 100%;
    }
```

And keep `aria-hidden="true"` on the `<div class="container">`

pennin-style:
@asyncLiz 

Thank you for your prompt feedback.

We will make adjustments based on the information you provided.

This is my first pull request, so please excuse my rudimentary question.

I looked at the `package.json` file to set up a development server, but there are no scripts such as `dev` or `start` set up, so I couldn't figure out how to start it.

It seems that `npm run build` will build the server, but will I have to load this script in a separate `index.html`?

We apologize for the inconvenience, but we would appreciate it if you could let us know.

asyncLiz:
We don't have a great dev server setup for external contributors yet. The only option right now is to [run the package's catalog server](https://github.com/material-components/material-web/discussions/5545) and navigate to `/components/radio/stories/`.

pennin-style:
@asyncLiz 
Thank you very much for your kind attention.
I was able to successfully set up the development server.

We will contact you again as soon as the adjustments are completed.

pennin-style:
@asyncLiz 

Please excuse me for repeating myself.

Is `<md-radio>` in `radio/demo/stories.ts` provided by unpkg?

I need to edit `radio/internal/radio.ts` this time, but I am having trouble getting the changes to take effect.

On the other hand, the elements in `radio/demo/stories.ts` reflect the changes (e.g. changing “Cats” to “Cat”).

Your wisdom would be greatly appreciated.
Thank you in advance.

asyncLiz:
> Is `<md-radio>` in `radio/demo/stories.ts` provided by unpkg?

It may be, that sounds familiar. I haven't tried running it locally for development :(

I'm not sure I have a good suggestion. I think @VandeurenGlenn mentioned they had a script in their fork of the repo for development.

pennin-style:
@asyncLiz

Thank you for your response.
I understand the content.

So for now, I will only send the changes you proposed and he will be responsible for confirming the operation status, is that correct?

asyncLiz:
You can make the changes and I'll verify it

asyncLiz (approved):
Thanks! I'll take a look at this and test it internally

radio/internal/_radio.scss
```diff
@@ -89,17 +89,13 @@ $_md-sys-motion: tokens.md-sys-motion-values();
...
-      margin: 0;
       position: absolute;
       width: 48px;
-      cursor: inherit;
     }
 
-    :host([touch-target='none']) input {
+    :host([touch-target='none']) .touch-target {
       width: 100%;
       height: 100%;
```
asyncLiz, line 100: Things look good! The one thing I see to improve is we can `display: none` the touch target instead of making it fit the width/height, like we do with [button's touch target](https://github.com/material-components/material-web/blob/22e19c264a56241d84dabeb399dbfc2ebc352a63/button/internal/_touch-target.scss#L21). Mind making that change?

```css
:host([touch-target='none']) .touch-target {
  display: none;
}
```

asyncLiz (changes requested):

radio/internal/_radio.scss
pennin-style: Understood. I will make the changes and get back to you shortly.

asyncLiz (approved):

pennin-style:
@asyncLiz 

Thank you very much for the series of guidance and reviews.

PR #5701, opened 2024-08-25
fix(labs): remove unnecessary CSS duplication variables of components in labs

bre97-web:
I found that some components in the labs folder had duplicate CSS variables (includes badge, navigationbar, navigationtab, segmentedbutton) , like this:

image

After I compile the component style separately through sass:

`npx sass ./labs/navigationbar/internal/navigation-bar-styles.scss ./o.css`

image

This is because of unnecessary prefixes in these styles:

```scss
    @each $token, $value in $tokens {
      // We need to delete "var(--md-navigation-bar-#{$token}, "
      //--_#{$token}: var(--md-navigation-bar-#{$token}, #{$value});
      --_#{$token}: #{$value};
    }
```

After repair:

image

### DIFF
--- a/labs/badge/internal/_badge.scss
+++ b/labs/badge/internal/_badge.scss
@@ -29,7 +29,7 @@
 
   :host {
     @each $token, $value in $tokens {
-      --_#{$token}: var(--md-badge-#{$token}, #{$value});
+      --_#{$token}: #{$value};
     }
   }
 
--- a/labs/navigationbar/internal/_navigation-bar.scss
+++ b/labs/navigationbar/internal/_navigation-bar.scss
@@ -35,7 +35,7 @@ $_md-sys-motion: tokens.md-sys-motion-values();
 
   :host {
     @each $token, $value in $tokens {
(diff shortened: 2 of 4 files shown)


asyncLiz:
This is working as intended. The custom properties with underscores are private internal details and not meant to be used directly. The prefixed component properties such as `--md-badge-*` are the public API surface.

bre97-web:
> This is working as intended. The custom properties with underscores are private internal details and not meant to be used directly. The prefixed component properties such as `--md-badge-*` are the public API surface.

Sorry to bother you. I still don’t understand why `--md-badge-*` appears twice.

e.g.
```css
--_size: var(--md-badge-size, var(--md-badge-size, 6px));
```

asyncLiz:
I'm sorry, I didn't pay enough attention. You're right, there should only be one `--md-*` var.

asyncLiz (approved):
Thanks for catching this, I missed it when adding the ability for tokens to return custom properties directly.

PR #5704, opened 2024-09-05
feat(list): add activatedItem getter

vdegenne:
Proposing this helper function to quickly retrieve the activated item in the list.

### DIFF
--- a/list/internal/list.ts
+++ b/list/internal/list.ts
@@ -90,4 +90,13 @@ export class List extends LitElement {
   activatePreviousItem(): ListItem | null {
     return this.listController.activatePreviousItem();
   }
+
+  /**
+   * Returns the activated item in the list.
+   *
+   * @return The activated list item or `null` if there are no items.
+   */
+  get activatedItem(): ListItem | null {
+    return this.items.find((i: ListItem) => i.tabIndex === 0);
+  }
 }


list/internal/list.ts
```diff
@@ -90,4 +90,14 @@ export class List extends LitElement {
...
     return this.listController.activatePreviousItem();
   }
+
+  /**
+   * Retrieves the first activated item of the array of items.
+   *
+   * @return A record of the first activated item including the item and the
+   *     index of the item or `null` if none are activated.
+   */
+  get activeItem() {
```
asyncLiz, line 100: Can you move this above next to the other `get items()` property and add `@export` to the jsdoc?

list/internal/list.ts
```diff
@@ -90,4 +90,14 @@ export class List extends LitElement {
   activatePreviousItem(): ListItem | null {
     return this.listController.activatePreviousItem();
   }
+
+  /**
+   * Retrieves the first activated item of the array of items.
```
asyncLiz, line 95: nit: "array of items, or null." to be more explicit

list/internal/list.ts
```diff
@@ -90,4 +90,14 @@ export class List extends LitElement {
...
+   *
+   * @return A record of the first activated item including the item and the
+   *     index of the item or `null` if none are activated.
+   */
+  get activeItem() {
+    return this.listController.activeItem;
```
asyncLiz, line 101: I don't think we should expose the `{item, index}` object that the controller uses. Instead, I would suggest separate `activeItem` and `activeItemIndex` properties, like [how tabs work](https://github.com/material-components/material-web/blob/a9ee4f5bc1d6702e5dc352eefed13a1d849577e3/tabs/internal/tabs.ts#L45-L99).

asyncLiz (changes requested):

PR #5710, opened 2024-09-09
fix(filter-chip): center trailing icon in its container

vdegenne:
fixes https://github.com/material-components/material-web/issues/5709

### DIFF
--- a/chips/internal/_trailing-icon.scss
+++ b/chips/internal/_trailing-icon.scss
@@ -30,8 +30,11 @@
   }
 
   .trailing.icon {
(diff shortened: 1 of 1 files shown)


vdegenne:
@asyncLiz This is a terrible fix, now I understand the problem, the icon is keeping its original size of 24px but trailing icon container is reduced to 18px. One workaround would be to set force the icon size from inside the trailing container so users don't need to implicitly write `--md-icon-size: 18px` every time they slot an icon there.

However I am not sure how to make this change.

PR #5714, opened 2024-09-13
Add missing overrides to symbol property methods

trevorade:
Without `override` on these, does not compile in TS 5.6


trevorade:
Doing this in google3 instead.

PR #5748, opened 2024-11-13
Add abstracted props in order to make internal components replaceable for all buttons

soelen:
So this idea is based on the interal select component class:

https://github.com/material-components/material-web/blob/main/select/internal/select.ts#L219
https://github.com/material-components/material-web/blob/main/select/internal/select.ts#L391-L417

with this feature it was possible for our team to extend this class and replace it with our customized filled field component;  similar to what we see in the filled select component:
 
https://github.com/material-components/material-web/blob/main/select/internal/filled-select.ts#L15

Our team faces right now the issue that the internal focus ring/ripple component is not switching its colors correctly when we switch the theme because these are wrapped around our own version of md-button/md-icon-button. Right now we apply hacky css based solutions in order to color them correctly.

With this PR we are able to customize all material web button components better, similar to select.

### DIFF
--- a/button/internal/button.ts
+++ b/button/internal/button.ts
@@ -4,9 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import '../../focus/md-focus-ring.js';
-import '../../ripple/ripple.js';
-
 import {html, isServer, LitElement, nothing} from 'lit';
 import {property, query, queryAssignedElements} from 'lit/decorators.js';
 
@@ -25,6 +22,7 @@ import {
   internals,
   mixinElementInternals,
 } from '../../labs/behaviors/element-internals.js';
+import {html as staticHtml, StaticValue} from 'lit/static-html.js';
 
 // Separate variable needed for closure.
 const buttonBaseClass = mixinDelegatesAria(mixinElementInternals(LitElement));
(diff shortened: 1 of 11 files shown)


asyncLiz:
We've discussed before how we might want to handle scoped custom element registration for elements used inside others (like `<md-ripple>`). We considered using lit static extensively, but there's a performance trade-off using lit static that we are not satisfied with, which is why we decided not to apply the approach to all subcomponents.

All subcomponents should be exposed for styling by a `::part()`. You can use that and part forwarding in your own wrappers to more easily style them.

soelen:
I understand the trade-off but unfortunately we have no much use for the `::part()` selector since we are extending our component classes from the internal component classes of material web (e.g. companyname-filled-button with companyname based color tokens etc).

@asyncLiz Before creating yet another PR I would like to ask in advance if following approach is maybe an option:

The iconbutton base class uses private marked methods in order to render md-ripple and md-focus-ring:

https://github.com/material-components/material-web/blob/main/iconbutton/internal/icon-button.ts#L222-L235

Is it an idea to mark these methods and protected and import these dependencies on an non internal level instead?

https://github.com/material-components/material-web/blob/main/iconbutton/internal/icon-button.ts#L7-L8

asyncLiz:
No, since making them protected will expose internal parts of our API that we don't intend.

Using internal code is not something we can scalably support. I'd instead recommend cloning the repo so that you can make more extensive changes in your own version.

PR #5758, opened 2024-12-15
fix(list): list item click() not working

vdegenne:
fixes https://github.com/material-components/material-web/issues/5705


asyncLiz (approved):
Sorry for the delay!

PR #5769, opened 2025-02-03
feat(button,chips,iconbutton): add download filename support for links

TonySpegel:
Adds support for specifying a filename when downloading a file using a link button.
Example usage:

```html
<md-filled-tonal-button href="#" download="example-file.pdf">
  Download File (PDF, 1.5 MB)
</md-filled-tonal-button>
```

### DIFF
--- a/button/internal/button.ts
+++ b/button/internal/button.ts
@@ -72,7 +72,7 @@ export abstract class Button extends buttonBaseClass implements FormSubmitter {
    * If not specified, the browser will determine a filename. 
    * This is only applicable when the button is used as a link (`href` is set).
    */
-  @property() downloadFilename = '';
+  @property() download = '';
 
   /**
    * Where to display the linked `href` URL for a link button. Common options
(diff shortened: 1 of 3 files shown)


button/internal/button.ts
```diff
@@ -67,6 +67,13 @@ export abstract class Button extends buttonBaseClass implements FormSubmitter {
    */
   @property() href = '';
 
+  /**
+   * The filename to use when downloading the linked resource. 
+   * If not specified, the browser will determine a filename. 
+   * This is only applicable when the button is used as a link (`href` is set).
+   */
+  @property() downloadFilename = '';
```
asyncLiz, line 75: Can you rename this to `download`, to match the [HTMLAnchorElement.download](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/download) property?

It would also be super helpful to include and forward this property for [link chips](https://github.com/material-components/material-web/blob/ac916578a26ea8b6b19ad2e9fdeb3f4970c2e993/chips/internal/assist-chip.ts#L51) and [link icon buttons](https://github.com/material-components/material-web/blob/ac916578a26ea8b6b19ad2e9fdeb3f4970c2e993/iconbutton/internal/icon-button.ts#L193).

asyncLiz (changes requested):

asyncLiz (approved):
Thank you!

aogzpa (commented):
LGTM!

PR #5790, opened 2025-04-02
fix(menu): output menu item theme values

twentyrogersc:
This follows other implementations (eg. [`menu/internal/_menu.scss#L27`](https://github.com/material-components/material-web/blob/c72650fa2e1810c5497ed2e1c19e46ca6cb185d3/menu/internal/_menu.scss#L27)).

Fixes https://github.com/material-components/material-web/issues/5789.


asyncLiz (approved):
Thanks for the fix!

Issue #5121, opened 2023-10-25
md-dialog: scrollbar Flash on opening


vdegenne:
### What is affected?

Component

### Description

The scrollbar flashes when the dialog opens even in the case the content is not scrollable.
It's particularly more apparent in dark mode.
See below playground for a demonstration of this issue.

### Reproduction

https://lit.dev/playground/#gist=825f3acbc05233df00a8cdb1bd9693c2

### Workaround

Could this issue be fixed by hidding the overflow on animation start?

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

@material/web@1.0.1

### Browser/OS/Node environment

Any browser/system

asyncLiz:
I'm not able to reproduce on Chrome macOS. Which browser/os environment are you seeing the issue? Can you upload a screen recording too?

christophe-g:
I can reproduce on Linux Mint 21.2 Chrome 118.0.5993.88.

I cannot reproduce on FF.

https://github.com/material-components/material-web/assets/1168053/871fb0f6-b29b-497c-a96c-fce37b3e8f1c

SummersRA:
I can reproduce on Windows with Edge.  No issue with FF.

vdegenne:
Firefox is doing great (you can even select text in the dialogs 😹)

SDBonhof:
Can reproduce this with Chrome M114 and M122 running through Electron V0.25.X and V0.29.X, respectively

shhdharmen:
I found an interim fix, adding `overflow: visible` to the `.scroller` seems to be fix the issue. Not sure what side-effects will it cause, but works fine for small dialogs.

```ts
function fixScrollFlash(mdDialog: MdDialog) {
    const shadowRoot = mdDialog.shadowRoot;
    if (shadowRoot) {
      const scroller = shadowRoot.querySelector<HTMLElement>(".scroller");
      if (scroller) {
        scroller.style.overflow = "visible";
      } else {
        console.log("scrollernot found");
      }
    } else {
      console.log("shadowRoote not found");
    }
  }
```

Basyras:
My workaround. Slightly inspired by @shhdharmen workaround.

Using these 2 events it will make sure the scrollbar will work as needed after animation is played and the `shadowRoot `children are present - so no check for `.scroller` shouldn't be needed.

> **open**: Dispatched when the dialog is opening before any animations.
> **opened**: Dispatched when the dialog has opened after any animations.


```
[...]

PR #5802, opened 2025-05-12
fix(dialog): scrollbar Flash on opening #5121

bipedal-eel:
This prevents the scrollbar from flashing (Chromium) as mentionend in #5121

### DIFF
--- a/dialog/internal/_dialog.scss
+++ b/dialog/internal/_dialog.scss
@@ -173,7 +173,7 @@
     // space from appearing on platforms that reserve scrollbar space.
     // Note: we only scroll vertically. Horizontal scrolling should be handled
     // by the content.
-    overflow-y: scroll;
+    overflow-y: auto;
   }
 
   .content {
@@ -185,6 +185,7 @@
     font-weight: map.get($tokens, 'supporting-text-weight');
     height: min-content; // Needed for Safari
     position: relative;
+    overflow-y: clip; // Needed for .scrollable .scroller
   }
 
   slot[name='content']::slotted(*) {


dialog/internal/_dialog.scss
```diff
@@ -173,7 +173,7 @@
     // space from appearing on platforms that reserve scrollbar space.
     // Note: we only scroll vertically. Horizontal scrolling should be handled
     // by the content.
-    overflow-y: scroll;
+    overflow-y: auto;
```
asyncLiz, line 176: If I understand correctly, this works by not _forcing_ the scrollbar to show (`auto` vs `scroll`) if we detect overflowing content in JS and add the `.scrollable` class?

What is the advantage to this solution as opposed to addressing the issue in JS that results in an inaccurate toggling of the `.scrollable` class?

For example, toggling the class after open animations have settled as suggested in https://github.com/material-components/material-web/issues/5121#issuecomment-2094905986.

dialog/internal/_dialog.scss
```diff
@@ -185,6 +185,7 @@
     font-weight: map.get($tokens, 'supporting-text-weight');
     height: min-content; // Needed for Safari
     position: relative;
+    overflow-y: clip; // Needed for .scrollable .scroller
```
asyncLiz, line 188: Why is `overflow-y: clip` vs the default `overflow-y: visible` needed for the scroller?

dialog/internal/_dialog.scss
bipedal-eel: Yes, you understand correctly.

I suppose ive always been drilled on doing css over js but I can see your point, especially since there are open and close methods already.

Will do this in js 👍

dialog/internal/_dialog.scss
bipedal-eel: Not too sure on that actually... .content overflows when scroller is set to auto, so this prevents that. This would make more sense in a .scrollable > .content but as mentioned above ill go do this in js

bipedal-eel:
i have found the underlying issue:
image

the screenshot is console logs in render():
```js
protected override render() {
    console.log("isOpen", this.isOpen)
    console.log("open", this.open)
    console.log("opening", this.isOpening)
    const scrollable =
      this.open && !(this.isAtScrollTop && this.isAtScrollBottom);
    console.log(scrollable)
// ...
}
```

Now scrollable is only set when !isOpening.
```js
const scrollable =
      !this.isOpening && this.open && !(this.isAtScrollTop && this.isAtScrollBottom);
```
I tested this a bit and as far as i can tell, this does not break anything.

If it does, an alternative soloution would be the comment you mentioned or setAttribute("opening", "") and doing:
```css
.scrollable .scroller:not([opening]) {
  overflow-y: scroll;
}
```

PR #5813, opened 2025-06-02
fix(catalog): hydrate CSR component pages

aogzpa:
# Description
This `{% if not ssrOnly %}` ... `{% else %}` ... `{% endif %}` is a conditional statement that checks if a variable named `ssrOnly` is `false`.
If it is `false` then the JavaScript is loaded and hydrated onto the components nested inside on an idle callback via the `<lit-island>` utility.

Essentially, a `<lit-island>` tag is a wrapper or a container designed to encapsulate other Lit components and control when and how they are hydrated (become interactive) on the client side.

This PR fixes the layout structure to provide hydration or not depending on the `ssrOnly` attribute.

### DIFF
--- a/catalog/site/_includes/layouts/components.html
+++ b/catalog/site/_includes/layouts/components.html
@@ -12,11 +12,13 @@
 
 {% block content %}
 {% if not ssrOnly %}
-<!-- Loads page JS on idle callback at src/hydration-entrypoints/components/md-file-name.ts -->
-<lit-island
-    on:idle
-    import="/js/hydration-entrypoints{{ page.filePathStem }}.js">
-</lit-island>
+  <!-- Loads page JS on idle callback at src/hydration-entrypoints/components/md-file-name.ts -->
+  <lit-island
+      on:idle
+      import="/js/hydration-entrypoints{{ page.filePathStem }}.js">
+    {{ content | mdMarkdown | safe }}
+  </lit-island>
+{% else %}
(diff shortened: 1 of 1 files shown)


asyncLiz:
I'm not super familiar with the catalog's setup. What issue is this fixing/how can I test it?

aogzpa:
> I'm not super familiar with the catalog's setup. What issue is this fixing/how can I test it?

My apologies for the initial PR description's alarmist tone; upon review, this fix isn't addressing an obvious critical bug across all pages. Instead, it's about correcting the `<is-land>` tag's usage to align with its intended design and optimization patterns.

We're currently using `<is-land>` without wrapping the components it's meant to hydrate. This is a deviation from its authors' conception. While this may not cause immediate, visible errors everywhere, it negates the tag's purpose and introduces potential issues in specific scenarios, such as when fallback content is specified or when dependent components rely on the parent `<is-land>`'s ready state. An empty wrapper provides no component to hydrate, defeating the "island" concept.
Refer to the official documentation for correct implementation:
 https://github.com/11ty/is-land/blob/main/demo%2Fdemo-lit.html

Furthermore, analyzing the source code (https://github.com/11ty/is-land/blob/main/is-land.js) reveals a crucial hydration optimization: nested `<is-land>| elements wait for their parent `<is-land>` to complete hydration before proceeding. This sequential hydration underscores why `<is-land>` must always serve as a wrapper for the components it manages, ensuring efficient and dependent "island" activation.

PR #5842, opened 2025-08-22
docs: update API docs

FrankGDak47:


asyncliz:
We have a bot that does this automatically (https://github.com/material-components/material-web/pull/5841)

PR #5844, opened 2025-08-29
feat(menu): Add support for download/rel attributes in menu-item

paladox:
Only works if using a link.

We use these attributes at gerrit. I found a workaround for now but preferably native support should be added to menu-item.


paladox:
Bump :)

cc @asyncliz

I liked to use this for Gerrit, having to workaround this by doing https://github.com/GerritCodeReview/gerrit/blob/2d98d58ac7debea921af7f839102f691d74ebdc0/polygerrit-ui/app/elements/shared/gr-dropdown/gr-dropdown.ts#L365

PR #5845, opened 2025-09-02
fix(radio): move root assignment to mirror hostDisconnected

austinw-fineart:
Fixes a regression caused by 688ab3c where calling hostConnected / hostDisconnected in rapid succession can cause root to become null. This regression can be experienced in Angular where content projection causes it to call hostConnected → hostDisconnected → hostConnected in rapid succession. I'm not familiar enough with Angular's internals to know why it needs to do this multiple times in a row but regardless the two lifecycles should be mirrored.


asyncliz:
Seems like a simple fix to me, I'll import to run tests. Is there any way to simulate Angular's behavior? Would directly calling the lifecycles in the order you describe be a way to test it?

asyncliz (approved):
LGTM! I'll see if there are any complications when testing

Issue #5830, opened 2025-07-23
Issue with md-menu on Safari


paladox:
### What is affected?

Component

### Description

I was comparing the behaviour from chrome to safari so I think this is a bug. Essentially when you use a mouse and you click, it opens the menu and triggers the focus ring on safari. On chrome it only happens with the keyboard.

https://github.com/user-attachments/assets/0aa16c5b-fccc-4a22-92af-9557c2e80a13

Chrome

https://github.com/user-attachments/assets/b3b6b8dd-ef5d-46d1-b6ef-faf52bae24d5

Safari

### Reproduction

Click on the menu to open it. 

### Workaround

There is really no workaround.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

2.3.0

### Browser/OS/Node environment

Browser: Safari Version 18.5 (20621.2.5.11.8) and Chrome Version 138.0.7204.158
OS Mac OS 15.5
Node Version 24.3.0 (although we're using node 20 with bazel using the rules nodejs)
NPM Version 11.4.2 (using yarn 1.22.19)

PR #5885, opened 2026-01-31
focus-ring: Workaround issue on safari

paladox:
* Fixes an issue with focus-ring being triggered when opening any element that has focus-ring.
* Fixes an issue with focus-ring not working anymore after opening the element that has focus-ring in this case for example menu. Once opened, click on any of the items. Then click off screen. Now re-open and use the keyboard, notice how focus ring isn't working.

This fix is only applied to safari as chrome works. I think it also works correctly in Firefox but I haven't tested. It's a ugly workaround but it fixes the issue I think.

Bug: Issue #5830


paladox:
@asyncliz are you able to review this please? :)

I tried to do the fix in a good way but found it pretty impossible (at least I'm unaware how to do this in any other way that looks nice).

paladox:
Had this hack https://github.com/GerritCodeReview/gerrit/blob/master/polygerrit-ui/app/elements/shared/gr-dropdown/gr-dropdown.ts#L206. The :hover check was supposed to fix it so when you first open with the mouse it didn't trigger focus-ring. But appears because of the hack, it didn't work (when I removed the hack). But it did fix the other issue.

PR #5949, opened 2026-04-17
fix(radio): dispatch input event on keyboard navigation

### LINKED ISSUE #5447: md-radio: check by keyboard doesn't trigger 'input' event
### What is affected?

Component

### Description

Checking radio in a radio group via keyboard arrows should trigger 'input' and 'change' event. `<md-radio>` dispatches 'change' event only: [code](https://github.com/material-components/material-web/blob/5f93c82546bbb428dcc8461b2e34cfe49178b0eb/radio/internal/single-selection-controller.ts#L228).

### Reproduction

Can be observed [here](https://stackblitz.com/edit/webpack-webpack-js-org-xwa6ut?file=dist%2Findex.html)

### Workaround

None

### Is this a regression?

### Affected versions

1.1.1

Browser: Chrome/121.0.0.0 Safari/537.36 Edg/121.0.0.0
OS: Windows NT 10.0; Win64; x64

akinshaywai:
## Problem

Fixes #5447

Native `<input type="radio">` fires both `input` **and** `change` events when a radio button is selected via arrow key navigation. `<md-radio>` only dispatches `change`, which means:

- React users miss keyboard selections because React's `onChange` maps to the DOM `input` event, not `change`
- Any listener specifically watching for `input` events is silently skipped on keyboard navigation

## Fix

In `SingleSelectionController.handleKeyDown`, fire an `InputEvent('input', { bubbles: true, composed: true })` before the existing `change` event, matching native `<input type="radio">` order and behavior:

- `composed: true` so the event crosses shadow DOM boundaries (same as native)
- `input` fires before `change` (same order as native)

A new test `dispatches an input event on user navigation` is added alongside the existing `dispatched a change event on user navigation` test.

## Testing

```
// Before fix – input handler never called on ArrowRight
radio.addEventListener('input', () => console.log('input fired'));

// After fix – fires on both click and arrow key navigation
```

### DIFF
--- a/radio/internal/single-selection-controller.ts
+++ b/radio/internal/single-selection-controller.ts
@@ -232,12 +232,17 @@ export class SingleSelectionController implements ReactiveController {
         }
       }
 
-      // The next sibling should be checked, focused and dispatch a change event
+      // The next sibling should be checked, focused and dispatch input and
+      // change events since the change is triggered by a user action.
+      // This matches native <input type="radio"> behavior.
       nextSibling.checked = true;
       nextSibling.tabIndex = 0;
       nextSibling.focus();
-      // Fire a change event since the change is triggered by a user action.
-      // This matches native <input type="radio"> behavior.
+      // Fire input before change, matching native <input type="radio"> order.
+      // input is composed so it crosses shadow DOM boundaries like native.
+      nextSibling.dispatchEvent(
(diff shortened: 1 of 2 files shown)


akinshaywai:
I have signed the Google Individual CLA. Could the CLA bot please rescan this PR? @google-cla

PR #5983, opened 2026-05-31
fix(radio): dispatch input event on keyboard navigation

ErnestHysa:
Fixes #5949

Dispatches an `InputEvent` before the `change` event when keyboard-navigating radio buttons with arrow keys, matching native `<input type='radio'>` behavior where both `input` and `change` events fire on user interaction.

## Changes

- Modified `SingleSelectionController.handleKeyDown` in `radio/internal/single-selection-controller.ts` to dispatch an `InputEvent` before the `change` event
- Added test in `radio/radio_test.ts` to verify `input` event is dispatched on keyboard navigation

## Testing

Added a new test case `dispatched an input event on user navigation` that verifies the `input` event is fired when using arrow keys to navigate radio buttons.

### DIFF
--- a/radio/internal/single-selection-controller.ts
+++ b/radio/internal/single-selection-controller.ts
@@ -236,8 +236,11 @@ export class SingleSelectionController implements ReactiveController {
       nextSibling.checked = true;
       nextSibling.tabIndex = 0;
       nextSibling.focus();
-      // Fire a change event since the change is triggered by a user action.
-      // This matches native <input type="radio"> behavior.
+      // Fire an input event first, then a change event since both are triggered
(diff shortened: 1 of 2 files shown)


ErnestHysa:
CLA flag noted — checking the linked Google account. The change itself is a small dispatch tweak in single-selection-controller.ts with a test in radio_test.ts; happy to address any review feedback once the CLA clears.

asyncliz:
@ErnestHysa it looks like the commit was made with the @ernest account, which hasn't signed the CLA. Can you either amend the commit using your other account with a forced update, or sign the CLA for that account?

asyncliz (approved):
Thanks for your patience! This LGTM

PR #6004, opened 2026-06-30
fix(dialog): layer scrim above selected tabs

### LINKED ISSUE #4948: Increase z-index for md-dialog scrim
The z-index is currently 1. Selected tabs in md-tab component also have z-index 1 and do not get covered by the md-dialog scrim.

Jaksenc:
## Problem

When an `md-dialog` is rendered before `md-tabs`, the dialog scrim can appear behind the selected tab. Both the dialog scrim and selected tab currently use `z-index: 1`, so DOM order can let the selected tab paint above the scrim.

## Change

Add `isolation: isolate` to the `<md-tabs>` host so selected tab internals stay inside the tabs stacking context instead of leaking above sibling components such as dialog scrims.

This keeps the selected tab's existing `z-index: 1` behavior local to tabs and leaves dialog scrim styling, dialog focus behavior, animations, and native `<dialog>` usage unchanged.

## Validation

- `npm ci`
- `npm run build`
- `/Users/jaksen/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin/node node_modules/@web/test-runner/dist/bin.js dialog/dialog_test.js tabs/tabs_test.js --config web-test-runner.config.js`
  - `dialog/dialog_test.js` and `tabs/tabs_test.js`: 71 passed, 0 failed
- `git diff --check`

Note: `npx wtr dialog/dialog_test.js --config web-test-runner.config.js` failed before running tests in my local shell under Node 26 with a `yargs` ESM/CJS loader error. The focused browser test above passed using the bundled Node 24 runtime.

## Risk

Low. The change adds a stacking context to the existing tabs host and keeps the selected tab's current `z-index: 1`. The regression test renders an active `md-primary-tab` with a dialog and verifies the tabs host isolates the selected tab stacking layer.

## Out of scope

- Adding a new `<md-scrim>` component
- Switching dialog to native `::backdrop`
- Adding a public z-index token
- Increasing the dialog scrim z-index

## Issue linkage

Closes #4948

### DIFF
--- a/dialog/dialog_test.ts
+++ b/dialog/dialog_test.ts
@@ -96,7 +96,7 @@ describe('<md-dialog>', () => {
       expect(dialogElement.open).toBeFalse();
     });
 
-    it('renders scrim above selected tabs', async () => {
+    it('renders scrim with selected tabs isolated', async () => {
       const {harness, root, scrimElement} = await setupTest();
       const tabs = document.createElement('md-tabs');
       const selectedTab = document.createElement('md-primary-tab');
@@ -109,11 +109,12 @@ describe('<md-dialog>', () => {
 
       await harness.element.show();
       const scrimZIndex = Number(getComputedStyle(scrimElement).zIndex);
+      const tabsIsolation = getComputedStyle(tabs).isolation;
       const selectedTabZIndex = Number(getComputedStyle(selectedTab).zIndex);
       expect(selectedTab.active).withContext('selected tab active').toBeTrue();
-      expect(scrimZIndex)
-        .withContext('scrim above selected tab')
-        .toBeGreaterThan(selectedTabZIndex);
+      expect(scrimZIndex).withContext('scrim z-index').toBe(1);
+      expect(selectedTabZIndex).withContext('selected tab z-index').toBe(1);
+      expect(tabsIsolation).withContext('tabs isolate child z-index').toBe('isolate');
     });
 
     it('fires open/close events', async () => {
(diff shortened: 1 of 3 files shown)


asyncliz:
Since arbitrary z-indices are difficult to maintain, can you instead fix this by adding [`isolation: isolate`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/isolation) to the [`:host` styles of `<md-tabs>`](https://github.com/material-components/material-web/blob/4413a6eca8273d5573613bd78c718ae2e7d9773b/tabs/internal/_tabs.scss#L10-L16)?

I already tested and confirmed this fixes the issue presented in #4948, and will be the preferred fix for other component z-indices 'leaking' out of their element.

Jaksenc:
Thanks, that makes sense. I updated the PR to use `isolation: isolate` on the `<md-tabs>` host instead of bumping the dialog scrim z-index.

That does feel like the cleaner fix: the selected tab can keep its existing `z-index: 1`, but that stacking now stays contained inside `<md-tabs>` instead of leaking above sibling components.

I also updated the regression coverage to render a dialog before an active `md-primary-tab` and verify that tabs are isolated.

Validation after the update:

- `npm run build`
- `/Users/jaksen/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin/node node_modules/@web/test-runner/dist/bin.js dialog/dialog_test.js tabs/tabs_test.js --config web-test-runner.config.js`
  - `dialog/dialog_test.js` and `tabs/tabs_test.js`: 71 passed, 0 failed
- `git diff --check origin/main..HEAD`

